PR #549, opened 2015-01-12
Network2

sjsrey:
compare allNeighborDistances with allNeighborDistances1 in the notebook. if speedup and logic is an improvement we can replace the former with the latter.


sjsrey:
should go into network2 from sjsrey not pysal/master (yet). will redo

PR #554, opened 2015-01-15
Beginning documentation cleanup

jlaura:
This is in reference to my documentation issue and is composed to changes to the documentation so that we are more machine parsable.

No need to merge yet, but I need to be able to pull the PR into the pysalREST account.


jlaura:
Going to reopen with a clean PR.

PR #555, opened 2015-01-18
Added edge segmentation by distance

jlaura:
Updates to the iPython notebook as well to illustrate segmentation


jlaura:
Merge into master is incorrect...

PR #560, opened 2015-01-26
modifying import scheme for network module

sjsrey:
Please look at the notebook and cells 9, 10 and 11 where the w for the network is returned rather than set as an attribute. the old code is commented out as it was raising an error. 

Also see the new import scheme in cells 1 and 2.


jlaura:
Import style looks good.  I must have missed adding the notebook before a PR after changing the W from an attribute to a returned object.  Good catch!

PR #571, opened 2015-02-04
Dict to unique value mapper

darribas:
Adding the option to pass a dictionary to explicitly assign colors to unique value mapper. This doesn't break previous API (as it stays as the default) but makes it possible to assign specific colors to unique value choropleths. Technically, it's all it's needed for easy and straightforward LISA mapping.

### DIFF
--- a/pysal/contrib/viz/mapping.py
+++ b/pysal/contrib/viz/mapping.py
@@ -389,7 +389,7 @@ def base_choropleth_classless(map_obj, values, cmap='Greys' ):
         map_obj.set_array(values)
     return map_obj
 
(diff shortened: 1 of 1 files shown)


darribas:
I didn't go ahead and create a wrapper for LISA because I wasn't sure of how to set it up (method in the `Moran_Local` class? independent method that takes a `Moran_Local` object and a shapefile?). Any ideas/suggestions on what you'd do?

For the record, a suitable color palette for LISA mapping can be this one:

```python
lisa_clrs = {1: '#FF0000', 2: '#66CCFF', 3: '#003399', 4: '#CD5C5C'}
```

PR #575, opened 2015-02-18
Network from Lattice

ljwolf:
Adds a function, `ntw_from_w` that constructs a `pysal.network,network.Network()` object from a lattice shapefile, using some spatial weights matrix. This allows us to port lattices into networks to solve problems with network structure like flow, assignment, or partition.

Currently, this arbitrarily assigns the centroids of the polygon to their dual network node.

This works best with contiguity weights, as the distances are computed from centroids to other centroids. I'd be curious on perspectives on how to change the edge distances given non-binary weights or using Luc's new spherical geometry code. Currently, this uses `numpy.linalg.norm`, so it's an easy swap.

### DIFF
--- a/pysal/network/floyd-warshall and repeat-dijkstra.ipynb
+++ b/pysal/network/floyd-warshall and repeat-dijkstra.ipynb
@@ -1,0 +1,275 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "import pysal as ps\n",
+    "import fw\n",
+    "import time\n",
+    "import numpy as np"
+   ]
+  },
+  {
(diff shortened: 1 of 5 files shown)


ljwolf:
And, if we're gonna keep the branch workflow, pysal/network or pysal/network2 needs to get updated.

sjsrey:
Should we leave this open, or is it ready for merge?

ljwolf:
yeah, let's merge this and update `pysal/pysal`'s network branch. That might be best.

sjsrey:
It looks like this one is to be merged into pysal/pysal:master?

We could make a clean network branch and merge this into that, if that was the intention?

ljwolf:
Yeah. Let's close this, make a new `pysal/pysal:network` that's level with master, and then I can do a separate PR. That way we have a `network` that starts level with `master` where network development for the next release can go.

That sound like a plan?

PR #584, opened 2015-02-24
Alpha in plot chor

darribas:
This allows to tweak transparency on the user method for a choropleth. It defaults to the previous set value so does not add additional complexity, only more flexibility.


darribas:
After merge, @sjsrey reported `mapping_guide.ipynb` throws an exception. Note to investigate it.

darribas:
This error turned out to be caused by something independent from the changes in the PR. In any case, it's been fixed now in #585

PR #592, opened 2015-03-14
adding check for version #591

pastephens:


jlaura:
The trusty build failed as expected due to the sphinx issue, but the failure on precise in the tutorials is unexpected.  It looks like an `import pysal as ps` statement is returning a string with the version name.  Is that the intended behavior?  If so, we need to update the tutorial / test.

sjsrey:
Version name has to be a string (1.9.1), so yes that is the intended behavior if that is what the question was about?

jlaura:
Check line 555 - 570:  https://travis-ci.org/pysal/pysal/jobs/54377037

The test is failing because of a return or a print.

pastephens:
Yeah, I'm surprised that there aren't many more of the same test failure
considering we import pysal a few times in the tests ;-)  I don't know what
to do about it. There are far too many tests to change, and the return
string will be changing all the time, too, so there's no point in modifying
the test suite to accommodate it.  Optimally, there would be a check in the
__init__ to see if a test suite (nose) had initiated the import (far-flung
idea probably). Thoughts?

pastephens:
The solution I came up with: created a new module, config.py, in wihich I store the toggle on/off value for checking for the most recent stable version of pysal, "check_stable". The main __init__.py imports config.py and obtains the value of this variable. In the travis.yml, I echo 'false' into config.py prior to running the test suite, so the new printed lines are suppressed, as well as the check of course. I also removed a print statement from travis.yml that printed the pysal version. This can now be our check that it works, as the check_stable should run in travis prior to the test suite/ echo rewrite.

pastephens:
I noticed that this really slows down the importation of pysal.  Alternatively, we could add info to version.py such as the release date, like Anaconda does, and some text saying we release twice a year in January and July.

sjsrey:
Added the tentative label so we can keep this open to discuss.

pastephens:
What's the preferred way to merge pysal/pysal upstream into my local
origin/591 branch?

jlaura:
git fetch upstream master
git checkout 591
git merge upstream/591

should do the trick

pastephens:
I had done:
git checkout master
git pull upstream master
git checkout 591

stephens1@pastephe-pc:~/Dropbox/pysal/pysal$ git merge upstream/591
merge: upstream/591 - not something we can merge

pastephens:
Travis build https://travis-ci.org/pysal/pysal/builds/55706615 is good, only Coveralls fails. This output is a bit verbose, but instructions for suppressing it are included inline. Otherwise, good to go.

pastephens:
This is ready to go. The implementation costs nothing on import if "check_stable" is set to False in config.py, which is achieved interactively at the prompt the first time the user imports a new version of pysal. Check it out, it's pretty cool!

jlaura:
Just retested via travis and this looks clean.  Not touching any of the PRs that have been merged in the interim, so I am merging.

PR #598, opened 2015-03-23
IO in Python 3

ljwolf:
This pull on FileIO passes all FileIO unittests in both python 2.7 and python 3.4, if converting using six fixers in `python-modernize` and `2to3`. I've rebased a whole bunch of working commits, so these are blocks of changes ordered around a theme.

If you want to play around with it, use the `mk3` script in the root directory. It'll make a new branch, 2to3, and convert the code, applying automatic `six` fixers, as well as rewriting the code in place, so make sure your branch `2to3` is expendable before running `mk3`. 

Otherwise, verifying this through other tests (and fixing nosetests for 3.4) will be the next step to a pysal in python3.


sjsrey:
On another PR #592 I added a `tentative` label for PR's that were intended for discussion but not yet to be merged. I wanted to check if this one should be merged (which I think is the intention) or if the tentative label should be applied.

ljwolf:
Yeah, I'd like to merge. No clue why coverage reduced, though.

Getting to 3 from here should be pretty straightforward, but anyone who wants to participate on that needs this IO stuff.

PR #602, opened 2015-03-25
Documentation Extraction Notebook

jlaura:


darribas:
Not sure if this is relevant in this context, but IPython 3 recently introduced the ability to run notebooks from the command line and check for discrepancies (use the `--to notebook` arg):

http://ipython.org/ipython-doc/stable/notebook/nbconvert.html

PR #622, opened 2015-06-12
New network branch from clean master

jlaura:
Initial commit in a PR to move network to be by geometry.  This add the 255KB New Haven, CT dataset.  No point data is included, yet.

  Issues to be tracked here:

* IO by geometry
* Point to chain snapping
* Any applicable cleanup.


sjsrey:
Tests are failing so I'm not going to merge until they get fixed. Also not sure how this relates to the larger pr #624 ?

sjsrey:
@jlaura and @dfolch we should have a short hangout to decide how to proceed with the network issues/prs.

Issue #459, opened 2014-06-17
Problem with bandwidth


bastienpoggi:
Hello,

I want to compute the connections between different nodes (coordinates) with different communication range (Connecity). I get the following error. Anybody knows why?
###############################################################
import pysal

coordinates = [ (3,4), (5,6), (50,1), (10,10) , (10,11), (5,3) ]
communicationRange =[25.0, 15.0, 25.0, 16.0, 14.5, 25.0]

connectivity = pysal.weights.Distance.Kernel(coordonnees,bandwidth=communicationRange)

print connectivite.neighbors
print connectivite.weights 

###############################################################
Traceback (most recent call last):
  File "D:/poggi_b/Desktop/Calcul de distances.py", line 6, in <module>
    connectivity = pysal.weights.Distance.Kernel(coordinates,bandwidth=bw)
  File "C:\Python27\lib\site-packages\pysal\weights\Distance.py", line 326, in __init__
    self._eval_kernel()
  File "C:\Python27\lib\site-packages\pysal\weights\Distance.py", line 380, in _eval_kernel
    zi = np.array([dict(zip(ni, di))[nid] for nid in nids]) / bw[i]
TypeError: zip argument #1 must support iteration

ljwolf:
@paulochf & @jlaura's suggestion keeps tests passing and solves @bastienpoggi's bug:

```
In [1]: import pysal

In [2]: coords = [ (3,4), (5,6), (50,1), (10,10) , (10,11), (5,3) ]

In [3]: commR = [25.0, 15.0, 25.0, 16.0, 14.5, 25.0]

In [4]: conn = pysal.weights.Distance.Kernel(coords, bandwidth=commR)

In [5]: print conn.neighbors
{0: [0, 1, 3, 4, 5], 1: [0, 1, 3, 4, 5], 2: [2], 3: [0, 1, 3, 4, 5], 4: [0, 1, 3, 4, 5], 5: [0, 1, 3, 4, 5]}

In [6]: print conn.weights
[...]

PR #626, opened 2015-06-13
cast arrays over inconsistent kdtree return types

ljwolf:
fixing #459 using simple typecheck


jlaura:
Is the indentation wrong in line 381 or is this an artifact of the git diff?

ljwolf:
Hmm... I think it's wrong on the diff [since it's not wrong in my repo](https://github.com/ljwolf/pysal/blob/arraycast/pysal/weights/Distance.py#L381). 

That's super weird tho. I can `:retab` and resub if it's worrying?

jlaura:
All tests passed, so I am guessing this is a git issues - it would have blown up otherwise.

PR #624, opened 2015-06-12
Network topology

dfolch:
This PR adds flexible functionality for dealing with typical topology problems described in #623. 

The creation of a Network object now allows the user to optionally remove duplicate road segments and apply minor rounding to shapefile node locations.  This facilitates the process simplifying the road network to a simple graph of intersections and endpoints.   Work done with @jGaboardi.

PR #627, opened 2015-06-14
Networkrb

sjsrey:
This cherry picks from #624.

However, while the unit tests pass, the doctests in https://github.com/pysal/pysal/blob/master/pysal/network/network.py will fail and need to be fixed. They evidently were not tested before the merge?

See https://github.com/pysal/pysal/blob/master/pysal/network/network.py#L253
`counts` isn't defined in the scope of that doctest. there are other failures as well.

We shouldn't merge this until

- we agree that we have the right commits here from #624 
- the doctests are fixed

### DIFF
--- a/pysal/network/network.py
+++ b/pysal/network/network.py
@@ -22,7 +22,16 @@ class Network:
     Parameters
     -----------
     in_shp : string
-             A topoligically correct input shapefile
+             input shapefile
+
+    node_sig : int
+               round the x and y coordinates of all nodes to node_sig 
+               significant digits (combined significant digits on left and right 
+               of decimal place); default is 11; set to None for no rounding
+
+    unique_segs : boolean
+                  if True (default), keep only unique segments (i.e., prune
+                  out any duplicated segments); if False keep all segments
 
     Attributes
     ----------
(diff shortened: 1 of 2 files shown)


dfolch:
Thanks for cherry picking... these are the only two commits I had intended for #624.

jlaura:
Working on doctests now.

jlaura:
Doc tests are fixed.  I pulled this PR locally to make the changes using:

git checkout -b sjsrey-networkrb master

git pull https://github.com/sjsrey/pysal.git networkrb

Any thoughts on how to update this PR from my side?

dfolch:
@jlaura If it will make life easier, we can reject this PR (and #624).  You can put the network doctest fixes into its own PR (or add them to #628), and then I'll resubmit the topology stuff under a new PR after pulling those fixes.

jlaura:
As a note - until the open PR for doc-tests is merged, it is going to be necessary to pull this PR locally and manually run the doctests to verify.

sjsrey:
@jlaura we could keep the doctests separate for now, rather than have multiple prs hanging around waiting for another one to get closed. So once you and @dfolch are ok with the network related commits in this pr, this could be merged. Then we can revisit what to do about doctests in the upstream/travis testing. One option for the latter would be to have devs run doctests locally before a pr (as well as the unit tests) but leave the doctests off in travis.

dfolch:
I think it would be safest to turn the doctests back on in travis (once everything is fixed).  I talked to @pedrovma and he is good with changes to spreg needed to get all the tests to pass. I'll submit a PR now to fix spreg.

@jlaura let me know if you have any questions about the topology stuff.  Here is a zoom-in on major roads in Rhode Island.  "Before" (without rounding and removing duplicate segments) and "after". The red dots are what pysal identifies as intersections, the blue lines are actual roads and the grey lines are graph edges.
none_false

11_true

PR #628, opened 2015-06-15
Re-enable doctests

jlaura:
The March14 fix for sphinx and travis-ci compatibility also killed doctests.  This PR re-enables doctests via Travis without making any alteration to Sphinx.

The linked Travis Build are the results from my master:  https://travis-ci.org/jlaura/pysal/jobs/66811136
This PR should also generate Travis-CI output with the culprits.

The network doctest failures are fixed in another PR.  The other failures will need attention, but perhaps on a different PR (some are in SPREG and look like rounding issues).

Others are in gamma.py and weights.py.


jlaura:
It sounds like we are going to discuss testing structure.  I am going to close this PR until we discuss testing options.

PR #629, opened 2015-06-17
changes to spreg tests for travis

dfolch:
This PR is intended to get spreg in line with travis if #628 is accepted. Everything in spreg now passes when I call ```nosetests --with-coverage --cover-package=pysal --with-doctest``` on my machine.  However, the stuff my machine said needed fixing is slightly different from the spreg errors seen here: https://travis-ci.org/jlaura/pysal/jobs/66811136.  I also had to fix some of the unit tests to get this to pass on my machine.  I'll be interested to see what travis thinks of these changes.

I talked to @pedrovma, and he will merge these changes to the SVN side in the future.

### DIFF
--- a/pysal/spreg/regimes.py
+++ b/pysal/spreg/regimes.py
@@ -70,7 +70,7 @@ class Chow:
     >>> print 'Joint test:'
     Joint test:
     >>> print olsr.chow.joint
-    (0.6339319928978806, 0.8886223520178802)
+    (0.63393199289787527, 0.88862235201788142)
     '''
 
     def __init__(self, reg):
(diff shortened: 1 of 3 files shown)


sjsrey:
Didn't we agree to only pull from spreg? i'm concerned that we now will have to make changes in two places and this is why we decided spreg in pysal/pysal was off limits. changes go into svn then pysal/pysal from svn?

If it is just to test how these changes would impact travis (which i agree is nice), I think you can set up the testing on your own fork of pysal.

dfolch:
According to @pedrovma they have always had trouble getting the tests to pass in both the SVN and git, so that part of the code base has had mismatches.  I think that is why he wasn't too concerned about these changes affecting the SVN side.  There is nothing substantive here, just some tweaks to the rounding to make travis happy.  I'll try to get travis working in my fork.

sjsrey:
It probably makes sense to have a small group meet to come up  with a concerted standard for handling the tests. This way we are all on the same page.

pedrovma:
I'd be up for a short meeting anytime to discuss this. It's been a while since we last discussed how to proceed with these issues. An update would be welcome. This PR was meant as a very specific fix to Travis results only, that's why @dfolch and I thought it would be okay to exceptionally invert the standard procedure. But you're right, just checking a fork also works.

Issue #635, opened 2015-06-19
simulating points on vertical lines


dfolch:
Simulating points on a network works in all cases except for those with vertical lines.  A divide by zero error arises when computing the slope on vertical segments.

jlaura:
Closed in PR636.

PR #636, opened 2015-06-19
vertical line point simulation

dfolch:
Addresses #635.  Allows points to be simulated on networks with vertical lines.


dfolch:
I don't see any problems in the coveralls output, but it's reporting an error here.  Does this have to do with two separate PRs on the same file?

Issue #621, opened 2015-06-11
network distance speedup


dfolch:
We ran a profiler on the following distance computation in network and got some interesting results...
```
import pysal as ps
import numpy as np
ntw = ps.Network('Waverly.shp')
np.random.seed(444)
npts = 100
sim = ntw.simulate_observations(npts)
dis = ntw.allneighbordistances(sim)
```
The output below shows that the ```generatetree()``` function takes most of the time. 

![with_generatetree](https://cloud.githubusercontent.com/assets/1322332/8106784/6189422e-1012-11e5-9e02-7282afbc7887.png)

However, when looking over the entire network module, the output of ```generatetree()``` is not currently being used elsewhere (or at least I'm not seeing it).  Removing the ```generatetree()``` calculation reduces the total time from 159 to 8.6 seconds, and test_network.py passes (except for the one checking if the ```generatetree()``` results are correct).

![without_generatetree](https://cloud.githubusercontent.com/assets/1322332/8107047/7a30d0c4-1014-11e5-879b-d9fd712ecd08.png)

jlaura:
Closed in PR641.

PR #637, opened 2015-06-19
speedup distance computations

dfolch:
This PR addresses #621.

By turning off ```generatetree()``` the time to compute a distance matrix is cut by approximately 90% (ymmv).  After turning this off, the code was tested using ```Network Usage.ipynb```.  The notebook tests most of the network module including NetworkG, NetworkK and Moran, and ran without problems.


jlaura:
(1) Tests are failing at line 103 because the tree is being used there.  It appears that the tree is not used in the code (only the test).  We should confirm that the test coverage is what we want even when the tree is not generated (I think it is).

(2) This PR needs to rebase from master after merging the snapping code.

dfolch:
1. Thanks for catching that test failure.  
 2. Is it enough to do a ```git pull upstream master``` into my local branch, or is something more sophisticated needed?

jlaura:
Git fetch upstream master
Git checkout master
Git merge upstream/master

dfolch:
I will reopen this under a new PR since my branch is a total mess.

PR #647, opened 2015-06-30
Addresses 646

### LINKED ISSUE #646: arc distance in knnW
I may be missing something, but it seems to me that knnW in Distance.py is assuming that the distances are Euclidean and does not allow for arc distances. In practice, this will likely seldom come up as a problem, since, e.g., knnW_from_array (in user.py) takes a radius argument and passes a distance_metric argument to pysal.cg.KDTree before calling the worker bee knnW function, thus passing the correct KDTree.

However, when knnW is called directly (in Distance.py) there doesn't seem to be a way to pass a distance_metric argument and when only an array of coordinates is provided, the KDTree is called with only one argument (u) - on line 122.

Does this need to be addressed or is it taken care of in another way? One way to deal with this is to restrict knnW to always take a KDTree as an argument and remove the option to pass an array. This would then require a small helper function that creates the tree and deals with the distance_metric argument.

jlaura:
fixes #646

### DIFF
--- a/pysal/cg/kdtree.py
+++ b/pysal/cg/kdtree.py
@@ -237,7 +237,7 @@ class Arc_KDTree(scipy.spatial.KDTree):
 
 
 def KDTree(data, leafsize=10, distance_metric='Euclidean', radius=1.0):
-    if distance_metric == 'Euclidean':
+    if distance_metric.lower() == 'euclidean':
         if int(scipy.version.version.split(".")[1]) < 12:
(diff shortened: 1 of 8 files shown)


jlaura:
I missed knn calls in edsa.  Do not merge yet,

jlaura:
Tests are passing locally and on the second travis run.  The first run is failing because of issues booting the machine and getting the environment set.  I am manually rerunning the tests now, via Travis.  Once passing, this is ready for review.

Issue #652, opened 2015-07-09
Use cKDtree for Arc_KDTree


sjsrey:
https://github.com/pysal/pysal/blob/master/pysal/cg/kdtree.py#L42

Issue #651, opened 2015-07-09
Design Point Pattern API


sjsrey:
For new module

sjsrey:
Work on this is being carried out in the [points_contrib](https://github.com/sjsrey/pysal/tree/points_contrib) branch.

ljwolf:
work on this is complete in [pointpats](https://github.com/pysal/pointpats).

PR #659, opened 2015-07-10
kdtree changes for arc distance

dfolch:
This PR applies issue #646 to the entire library, and extends PR #647. It also addresses #652 and does some other stuff.

This PR allows the user to choose euclidean or arc distance (via the kd-tree) when using distance related functions. It does not change the API or reduce documented functionality. However, we might want to think about point handling more broadly in the future (see #651). It also speeds up arc distance queries by using cKDTree under the hood.

Changes:

 1. The default functionality for most distance related functions and classes is to take a kd-tree instead of an array of points. When an array is passed, we convert to euclidean distance kd-tree and give a deprecation warning. The only functions and classes that don't require a kd-tree are those that explicitly take an array or shapefile. 

 2. Made the pysal kd-tree accessible at the top level, i.e. pysal.KDTree.

 3. Modified Arc_KDTree to be built on top of scipy.spatial.cKDTree instead of scipy.spatial.KDTree when the user has scipy 0.12 or greater.

Notes:

 1. None of the numeric values in the tests had to change for this PR to pass (on my computer).

 2. Previously, most of these functions accepted lists, even though the docs all called for arrays. Now, if an array is passed it raises a deprecation  warning and a list fails. This is an artifact of how the changes were implemented, not an intended consequence.

### DIFF
--- a/pysal/__init__.py
+++ b/pysal/__init__.py
@@ -83,6 +83,7 @@ from pysal.core.util.weight_converter import weight_convert
 import pysal.spreg
 import pysal.examples
 from pysal.network.network import Network, NetworkG, NetworkK, NetworkF
+from pysal.cg.kdtree import KDTree
 
 
 # Load the IOHandlers
--- a/pysal/cg/kdtree.py
+++ b/pysal/cg/kdtree.py
@@ -9,19 +9,67 @@ import scipy.spatial
 import numpy
 from scipy import inf
 import sphere
+from sphere import RADIUS_EARTH_KM
 
 __author__ = "Charles R Schmidt <schmidtc@gmail.com>"
+
+__all__ = ["DISTANCE_METRICS", "FLOAT_EPS", "KDTree"]
+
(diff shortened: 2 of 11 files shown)


ljwolf:
In hopes of getting this merged, I've tried to rebase these changes onto the current pysal/dev branch. It's located in my [ckdtree_gc_dist_rebase](https://github.com/ljwolf/pysal/tree/ckdtree_gc_dist_REBASE) branch, and I think I got all the essentials in this PR.

However, I had to conduct some manual merges to incorporate mutual changes. Plus, there are over 250 outstanding commits... So, I can't just PR it to @dfolch and let github take care of the rest :disappointed: 

If and when [travis tests close successfully](https://travis-ci.org/ljwolf/pysal/builds/110305951), I can submit this in the same fashion as #744. 

If only github had an automatic way of tagging something overdue for a rebase...

Any thoughts?

ljwolf:
merging changes over in ljwolf/pysal/tree/kdtree

Issue #653, opened 2015-07-09
network is returning NAN's on diagonal of distance matrix


sjsrey:
This causes downstream problems with the d mat is used in other programs.

ljwolf:
What downstream issues were reported? via discussion in #660 I think this might be a design decision. 

Showing users a masked distance matrix could be helpful, if they need zeros along the diagonal.

jlaura:
This flew totally under my radar. Have to agree that 0 on the diagonal is
not ideal.  It isn't that NaN is necessarily more accurate (0 is), but NaN
offers more flexibility.

ljwolf:
closing via #660

PR #660, opened 2015-07-10
insert zeros on symmetric matrix diagonal

dfolch:
This PR addresses #653.  This change inserts zeros on the diagonal of a distance matrix instead of  NANs.

### DIFF
--- a/pysal/network/network.py
+++ b/pysal/network/network.py
@@ -739,8 +739,8 @@ class Network:
                     nearest[p2,p1] = nearest[p1,p2]                    
         if symmetric:
             # populate the main diagonal when symmetric
-            #np.fill_diagonal(nearest, 0)
-            np.fill_diagonal(nearest, np.nan)
+            np.fill_diagonal(nearest, 0)
+            #np.fill_diagonal(nearest, np.nan)
         return nearest
 
     def nearestneighbordistances(self, sourcepattern, destpattern=None):


sjsrey:
@jlaura should have a look as I don't remember the original motivation for the nan fill?

jGaboardi:
As far as I can tell there is no benefit  to have NaN instead of float(0.0) on a distance-based matrix.

- J





On Jul 10, 2015, at 4:38 PM, Sergio Rey <notifications@github.com<mailto:notifications@github.com>> wrote:


@jlaura<https://github.com/jlaura> should have a look as I don't remember the original motivation for the nan fill?

ljwolf:
I can think of one for generic distance matrices. 

Imagine where we make a network for potential [spatial interaction situations](https://github.com/pysal/pysal/tree/master/pysal/contrib/spint) where a point is a valid origin but not a destination. 

In that case, the self-distance should not be 0, but `np.nan`.  Right now, if we make them `nan`, we can't remask from zero. But, if `nan`, it's simple to use [builtin masking tools](http://docs.scipy.org/doc/numpy/reference/generated/numpy.nan_to_num.html) to make `nan`'s zeros.

In general distance matrices, we need to preserve a way to differentiate nonsymmetric origin-destinations. This commit does this. But, we need to remember this.

If zero masking is currently burning someone's use case, we should totally merge. This commit only affects **symmetric** matrices, which is the point of using zero and not nan. For a *symmetric* matrix, @jGaboardi is def right.

ljwolf:
Also, was thinking about this the other day. For some distance matrix `dist`, it's *very* convenient to be able to call things like `dist.min(axis=1)` or `np.argmin(dist, axis=1)` to extract nearest neighbors from a pairwise distance matrix. Infilling diagonals with zeros removes this ability. Not sure if this is desirable.

dfolch:
These are good arguments for nan on the diagonal. I think that the rest of pysal assumes zeros on the diagonal, unless explicitly stated otherwise. My interpretation of the earlier design decision was that most of the downstream functions were expecting zeros (or some numeric value) in all slots, so that was the default. The function ```weights.util.insert_diagonal()``` was written for the non-typical cases. Sounds like the typical (or better) case for network is nan.

PR #663, opened 2015-07-13
Examples

sjsrey:
This adds a couple of utility functions to expose the built-in datasets (See below).

screenshot 2015-07-13 16 55 28


ljwolf:
I could write a test for this that checks if all of the examples at least show a short title? That would cause the coveralls to pass.

sjsrey:
Adding tests would be great.

PR #664, opened 2015-07-14
Lisa map

darribas:
This PR introduces mid-layer and high-level methods to easily create LISA plots. A notebook showing how to use them to follow soon.


darribas:
**NOTE**: do not merge yet as some commits, in particular fb741f5 and 71c678f are not part of the PR.

darribas:
I think applying `git revert`, as suggested by @ljwolf has worked and the PR contains now what it was meant to. Should be ready to merge.

PR #678, opened 2015-07-28
W stitch

darribas:
This PR adds functionality in `Wsets` to stack weights objects and to created stacked objects where observations are connected also with its contemporary neighbors a number of periods backwards and forwards.

The methods are well documented and examples are provided but no unit tests are included. Also, I'm not sure what the best way to publicise this would be (beyond the docstrings), a companion notebook? If so, where should it be hosted?

This PR is related to functionality mentioned in #378, although it is not on-the-fly Ws.

### DIFF
--- a/pysal/weights/Wsets.py
+++ b/pysal/weights/Wsets.py
@@ -592,7 +592,8 @@ def w_stack(ws, silent_island_warning=False):
         for el in w.neighbors:
             out_neigh['%i-%s'%(i, str(el))] = ['%i-%s'%(i, str(j)) \
                                             for j in w.neighbors[el]]
-            out_weigh['%i-%s'%(i, str(el))] = w.weights[el]
+            out_weigh['%i-%s'%(i, str(el))] = []
+            out_weigh['%i-%s'%(i, str(el))] += w.weights[el]
         wid = ['%i-%s'%(i, str(j)) for j in w.id_order]
         out_ids.extend(wid)
     outW = pysal.W(out_neigh, out_weigh, id_order=out_ids, \
@@ -705,7 +706,8 @@ def w_stitch(ws, back=0, forth=0, silent_island_warning=False):
             # Contemporary neighbors
             out_neigh['%i-%s'%(i, str(el))] = ['%i-%s'%(i, str(j)) \
                                             for j in w.neighbors[el]]
-            out_weigh['%i-%s'%(i, str(el))] = w.weights[el]
+            out_weigh['%i-%s'%(i, str(el))] = []
+            out_weigh['%i-%s'%(i, str(el))] += w.weights[el]
             # Backward neighbors
             for t in range(1, back+1):
                 if i-t in range(len(ws)):
@@ -827,7 +829,8 @@ def w_stitch_single(w, t, back=0, forth=0, silent_island_warning=False):
             # Contemporary neighbors
             out_neigh['%i-%s'%(i, str(el))] = ['%i-%s'%(i, str(j)) \
                                             for j in w.neighbors[el]]
-            out_weigh['%i-%s'%(i, str(el))] = w.weights[el]
+            out_weigh['%i-%s'%(i, str(el))] = []
+            out_weigh['%i-%s'%(i, str(el))] += w.weights[el]
             # Backward neighbors
             for tb in range(1, back+1):
                 if i-tb in range(t):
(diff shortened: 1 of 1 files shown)


darribas:
@sjsrey Would this make sense for the new release? Don't think so, but if you do, I'm happy to push unit tests before release.

sjsrey:
I've been working on the release and really don't want to add any changes at this point (we always talk about the "freeze" but never really get there...).  

This looks very cool. I do think we might want to revisit where it goes in the library and how it may related to the spatial_dynamics stuff as well as spreg. So holding off till after the release for this merge would be what I vote for.

lanselin:
this will be great for the panel functionality we plan to have in spreg for the
jan 2016 release, not this one :-)


--  
Luc Anselin, PhD
Regents' Professor and Walter Isard Chair
Director, GeoDa Center for Geospatial Analysis and Computation
School of Geographical Sciences and Urban Planning
Arizona State University

darribas:
This sounds good, let's let it sit until post-release and then discuss where it goes best. I did feel `Wsets` might not be the best place as it's not really a "set" operation, but I do think it's really a `W` utility, so keeping it with other weights tools might make sense. In any case, food for thought, but to be eaten after this upcoming release.

pysal/weights/Wsets.py
```diff
@@ -531,3 +531,326 @@ def w_clip(w1, w2, outSP=True, silent_island_warning=False):
...
+    w                       : W
+                              Resulting `ps.W` object
+
+    Notes
+    -----
+    The resulting `w` contains the original indices, converted to strings if
+    necessary and preceded by 'X-', where X is the order of the original `W`
+    object in `ws`.
+
+    IMPORTANT: Weights are copied from the original weights object and do not
```
ljwolf: This should probably be a warning or exception:
```python
if ws.transform is 'O':
    raise Exception('Transformed weights cannot be passed')
```

pysal/weights/Wsets.py
```diff
@@ -531,3 +531,326 @@ def w_clip(w1, w2, outSP=True, silent_island_warning=False):
     return wc
 
 
+def w_stack(ws, silent_island_warning=False):
```
ljwolf: I think we might want to make the `ws` argument spatted?

```python
def w_stack(ws*, silent_island_warning=False): 
```
which would mean that instead of 
```python
>>> w_stack([W1, W2, W3], silent_island_warning=False)
```
we get
```python
>>> w_stack(W1, W2, W3, silent_island_warning=False)
```

pysal/weights/Wsets.py
darribas: Good catch!

Or de-standardize them ourselves first thing on the method?

pysal/weights/Wsets.py
darribas: Do we have a standard policy on this in `PySAL`? Should we if not?

pysal/weights/Wsets.py
ljwolf: Maybe? I'm not sure if it's that simple, since the [documentation warns about chaining transformations](https://github.com/pysal/pysal/blob/29f85b8c74053d40ea10533094a7ccfce00d8f5f/pysal/weights/weights.py#L752). 

But, practically, I can do the following just fine:
```python
kw0 =  ps.kernelW_from_shapefile(ps.examples.get_path('columbus.shp'))
kw = ps.kernelW_from_shapefile(ps.examples.get_path('columbus.shp'))
kw.transform = 'r'
kw.transform = 'o'
np.testing.assert_allclose(kw0.full()[0] , kw.full()[0])
```

pysal/weights/Wsets.py
ljwolf: No clue if it's used anywhere else in the library, but I used this pattern in `by_col_array`:
```python
filehandler.by_col_array('HOVAL', 'INC', 'CRIME')
# same as 
filehandler.by_col_array(['HOVAL', 'INC', 'CRIME']) 
```

sjsrey:
Should we move this over to libpysal.weights ?

darribas:
yeah, I think that'd be a good place. Not sure we decided where it'd be the best fit within `weights`. Any ideas? 

Thinking also about the [`rasterW`](https://github.com/pysal/libpysal/pull/150), it might be good to have a larger discussion on where these pieces could fit before they grow an eco-system around them (potentially) and it's messier to move. Tbh, I'm not sure exactly where I'd put it within `weights`, but definitely in `weights`.

sjsrey:
Closing as issue is now https://github.com/pysal/libpysal/issues/361

Issue #685, opened 2015-08-28
BV Lisa 


itsakifa:
I'm using BV_Moran and Univariate LISA. Is there a BV LISA tool planned ? This would be of great help!

Thanks and keep up the good work!
Isabel

darribas:
Hi Isabel,

Thanks for the message, I've labelled it as a feature request. My tentative answer is that it is not very hard to impement, but we'll discuss it in the developer meeting.

itsakifa:
Perfect, thank you!

darribas:
@itsakifa PR #686 has been merged into master so, if you're working on bleeding edge pysal (off the github repository's master branch), you can pull it in. It's documented but, essentially, you call it with `ps.Moran_Local_BV`.

PR #669, opened 2015-07-17
add nonsudo travis

ljwolf:
Modify the travis file to not use sudo.

Issue #491, opened 2014-07-19
Test np.allclose() for unit tests


pastephens:
Serge found said function and is testing for future ease of testing.

pastephens:
Serge can you provide a 2-line example of the change that needs to occur for this to come about?

darribas:
Bumped to next release as considered non-critical for this release.

PR #686, opened 2015-08-28
Moran local bivariate

darribas:
This PR addresses feature request #685 by introducing a class named `Moran_Local_BV`. The functionality has been tested against GeoDa, matching results when not dependent on randomization. The naming follows convention in the `moran` module and so does the API, which is a blend between `Moran_BV` and `Moran_Local`.

### DIFF
--- a/pysal/esda/moran.py
+++ b/pysal/esda/moran.py
@@ -2,7 +2,8 @@
 Moran's I Spatial Autocorrelation Statistics
 
 """
-__author__ = "Sergio J. Rey <srey@asu.edu>"
+__author__ = "Sergio J. Rey <srey@asu.edu>, \
+        Dani Arribas-Bel <daniel.arribas.bel@gmail.com>"
 from pysal.weights.spatial_lag import lag_spatial as slag
 from pysal.esda.smoothing import assuncao_rate
 import scipy.stats as stats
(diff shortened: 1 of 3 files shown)


ljwolf:
Travis is failing on array comparison in spreg's GMCombo tests. 

The recent change to miniconda and travis's new container-based architecture (#669) might cause this to happen more often unless we start using `np.allclose` in these cases.

darribas:
Those errors I think are unrelated to the PR though, which implements a new class that is not used anywhere else, definitely not in `spreg`.

pysal/esda/moran.py
```diff
@@ -697,6 +698,191 @@ def __quads(self):
...
+                   for two-sided tests, these values should be multiplied by 2
+
+    Examples
+    --------
+    >>> import pysal as ps
+    >>> import numpy as np
+    >>> np.random.seed(10)
+    >>> w = ps.open(ps.examples.get_path("sids2.gal")).read()
+    >>> f = ps.open(ps.examples.get_path("sids2.dbf"))
+    >>> x = np.array(f.by_col['SIDR79'])
```
ljwolf: neat trick. never realized that the `.by_col` admitted a `__get_item__`. Any special reason you're using this instead of `by_col_array()`?

ljwolf:
I think this should get merged.

'But, we need to *really* need to permanently address uniformity around `allclose` (#491) so we're not merging against failing but unrelated tests...

sjsrey:
-1


I agree this is a good addition and not the reason the tests are failing, but by merging we now have a broken build. Something we have tried to avoid. 

As much as a pain it is, I think we need to fix the problem rather than kicking the can down the road.

ljwolf:
PR incoming on the combo tests. I wasn't planning on merging this without merging a fix to the tests shortly. It's only 4 tests, and I have the fixes from the 2to3 conversion project.

pysal/esda/moran.py
darribas: I'm pretty sure I copied that logic from the docstring in `Moran_Local` :-) `by_col_array()` actually looks more sensible, I didn't know we had that!

darribas:
@ljwolf 's changes should address the errors and the build passes so we're back to a clean build, plus with the bivariate moran in master.

ljwolf:
I will continue to be working out the testing issues, implementing a consistent style and keeping with the preliminary work in #491 in the [allcloser branch](https://github.com/ljwolf/pysal/tree/allcloser). 

Sorry for merging these backwards. It's important to be careful with these things, and I know that. But, since I'd had the fixes in commits for the 2to3 branch already, I knew I just needed to cherry-pick them out.

PR #687, opened 2015-08-28
Change array assertions into allclose

ljwolf:
This changes some testing to `np.testing.allclose` as discussed in #491


ljwolf:
Fairly certain that, even with a switch to `allclose` in GMCombo, an error will be thrown because the last digit of an `assertAlmostEquals` will fail.

ljwolf:
This should now fix the blocking testing errors referenced in #686.

sjsrey:
+1
:-)

PR #690, opened 2015-09-05
fix for lisa markov one off for significance indicator

sjsrey:


ljwolf:
pernicious little bug!

PR #694, opened 2015-09-11
Allclose in SPREG

ljwolf:
For anyone interested, this is what it looks like if we swap 

`assert_array_almost_equal` and `assertAlmostEqual` to using `numpy.testing.assert_allclose`, with a relative tolerance variable stored in `pysal.common`. 

This pull addresses #491 and #687 

Nearly all tests pass at a relative tolerance of 1e-5. The few that do not require use of 1e-4. So, that's a precision of +/-1 every 10k. 

Some issues in tsls remain unresolved, and are likely to not be resolvable using relative tolerance alone, since the values being compared are O(1e-16), and we're comparing them to zero.

### DIFF
--- a/pysal/common.py
+++ b/pysal/common.py
@@ -16,7 +16,7 @@ except:
     print 'scipy 0.7+ is required'
     raise
 
-RTOL = .0000001
+RTOL = .00001
 
 import copy
 import math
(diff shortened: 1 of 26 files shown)


ljwolf:
closing & registering the branch in the wiki

PR #703, opened 2015-10-02
Swapping to Allclose and RTOL=.00005 in spreg

ljwolf:
Testing the swap to allclose and a relative tolerance of 1e-5 for spreg tests. Also, tightening the relaxation of the tests introduced in #694 

We can start the target on dev just in case, and syncronize to master and private spreg if this passes.

### DIFF
--- a/pysal/common.py
+++ b/pysal/common.py
@@ -16,7 +16,7 @@ except:
     print 'scipy 0.7+ is required'
     raise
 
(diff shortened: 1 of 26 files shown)


sjsrey:
Did we want this to go to master?

ljwolf:
submitting now

--
ljw

PR #710, opened 2015-10-22
Patsy/Pandas wrapper

ljwolf:
This provides a wrapper for PySAL regressions that, among other things, provides a Patsy/Pandas interface. Please check out [`Explaining the Handler`](https://github.com/ljwolf/pysal/blob/6de0b6617a28a2d99b88aa654f0ee4cd9ea4e9fe/pysal/contrib/handler/Explaining%20the%20Handler.ipynb) for a discussion of how it works. 

It belongs in `contrib` for a few reasons

1. It's *totally* optional. 
2. It introduces soft dependencies at the module level. 
3. By adding new endpoints to the registry, it will be useful to other modules as well, like `spatial_dynamics` or `esda`.


ljwolf:
Also, the `tests` in the module are current and work. But, in `contrib`, they won't get run by travis :P

datnamer:
Sweet :)

sjsrey:
This is really nice as I can see it adding nice interoperability between PySAL and many other packages.

darribas:
This is super cool. As we make the library more interoperable with others, such as `pandas`, we should think of ways to make it consistent. I'm not sure if the wrapper does this already but, for example, it'd be great if we had decorators that check if numpy arrays or pandas objects are being passed to make sure the returned output is of the same type. For example, we currently support passing a `pd.Series` object for the calculation of a spatial lag, but we still return a numpy array.

As I said, not sure this fits in this one, but something to keep in mind to implement at some point.

ljwolf:
Sweet, glad it's looking useful :)

I think type consistency is important as well.  I wonder, would that look like when we're dealing with large complex objects like a regression result?

Do we convert things in the formula back into pandas objects after estimation? Or store the original dataframe and have X, Y, etc. as views onto it? My chief initial concern was data duplication,  since I thought it was important to store the originating pysal model inside `model._called`, but hacking these into the handler now would be simple. 

I think for small things,  this "special dispatch based on input types" is very clearly addressed by multiple dispatch. Patterns/decorators for it are [easy to implement in python](https://en.wikipedia.org/wiki/Multiple_dispatch#Python).

I've really liked how it simplifies code like this in julia. For instance, if you had some function over scalars `a`,`b`,`c`:

```
def foo(a,b,c):
    return a*b + c
```

Then you could make it a [universal function](http://docs.scipy.org/doc/numpy/reference/ufuncs.html) by adding a method that calls `foo` recursively on the scalar contents of the lists:
```
def foo(alist,blist,clist):
    return [foo(a,b,c) for a,b,c in zip(alist, blist, clist)]
```
If we used a multiple dispatch strategy in pysal, I could see the current classes being the "base" class, and then other dispatches become wrappers around those functions, but use the same names. I don't know about speed costs on that in Python, though.

datnamer:
Theres a pretty robust multiple dispatch library here: https://github.com/mrocklin/multipledispatch/ 

Maybe  @mrocklin  can comment on the speed overhead and if it can be useful for this case.

mrocklin:
The per-call overhead of multipledispatch a small factor times the cost of a function call, somewhere in the microsecond range (if I recall correctly).  This makes it appropriate for functions that operate in the tens of microseconds or longer range but not for operations like `add`.

There is a non-trivial setup cost if you have hundreds of type signatures per function.  This is not as far-fetched as one might think.  This could be reduced with a little bit of effort though.

PR #714, opened 2015-10-28
Dev

sjsrey:
Keeping dev in-line with master until we move to rolling releases.


darribas:
Woops! Should have I issued PR #713 into dev instead of master?

sjsrey:
Well we are not really starting the following release scheme until January's release so no biggie.

Issue #717, opened 2015-11-05
esda.mapclassify return problematic counts when there is 0 occurrence in the last class


weikang9009:
![classify](https://cloud.githubusercontent.com/assets/7359284/10956421/b9dacf46-8318-11e5-8a64-5fc9096dae64.png)

The returned counts should be array([7, 4, 9, 0])

I think the problem lies in how numpy.bincount works. It only counts the occurrences in each bin up to the largest "binId", resulting in the the ignorance of the last bin if there is 0 occurrence.

ljwolf:
Yep, definitely weird... thanks for replication code. 

[This patch](https://github.com/pysal/pysal/compare/dev...ljwolf:bincount) fixes the issue. We can use `bincount`'s `minlength` keyword argument to force the vector of counts to be the same length as the vector of bins. This would pad the ending with zeros. 

Not sure if this is what we need in general, though...

weikang9009:
This is great! I think it well fixes the problem. It can return the right number even if there are more than 1 empty class.

ljwolf:
I wonder if we need this kind of testing anywhere else in `ESDA`, even in places where `bincount` isn't used... 

`User_Defined` the only one where you saw this occur?

ljwolf:
resolved by #718. Will reopen if other classifiers not using `bincount` have the same issue.

PR #718, opened 2015-11-05
force counts to be same length as bins

ljwolf:
This forces `User_Defined` to provide bin counts for all bins, and not omit zero-count bins at the tail of the distribution.


ljwolf:
Also, I should add, resolves #717 by @weikang9009

PR #720, opened 2015-11-20
Moving natural breaks to a cleaner kmeans implementation

sjsrey:


pysal/esda/mapclassify.py
```diff
@@ -273,38 +304,12 @@ def natural_breaks(values, k=5, itmax=100):
...
-            solved = True
-        else:
-            c0 = c1
-        it += 1
-        if it == itmax:
-            solving = False
-    class_ids = c1
-    cuts = [max(values[c1 == c]) for c in rk]
-    return sids, seeds, diffs, class_ids, solved, it, cuts
+    kres = _kmeans(values, k)
```
ljwolf: A real neat improvement to splatting in Python 3 could make this:

`class_ids, cuts, *_, fit, sids = _kmeans(values, k)`.

Unfortunately, this would be a syntax error in python 2, so this looks like the best way to do this.

pysal/esda/mapclassify.py
```diff
@@ -1093,16 +1095,29 @@ def _set_bins(self):
...
-        fit = res0[2].sum()
-        for i in xrange(self.initial):
-            res = natural_breaks(x, k)
-            fit_i = res[2].sum()
-            if fit_i < fit:
-                res0 = res
-        self.bins = np.array(res0[-1])
-        self.k = len(self.bins)
-        self.iterations = res0[-2]
+        values = np.array(x)
```
ljwolf: Would it be worth it in the future to break this off into a well-defined method, like `_can_classify(self, n)`? 

I imagine consistency checks like this logic may need to exist in `_set_bins` for each classifier (e.g. in quantiles if `k > np.unique(data)`).

ljwolf:
Looks good. Ready to go, or are there more commits on the way?

sjsrey:
Ready to go

ljwolf:
ok, looks like it's better to comment below relevant sections rather than at the start of relevant sections... :frowning:

PR #724, opened 2015-12-08
add synchronization tool

ljwolf:
I'm sure this could be done with sufficiently complex `awk`/`sed`, but that notwithstanding...

 I've written a python command line script that synchronizes a source and a destination, using only the files from `git ls-files` in the destination. This is a first stab at simplifying the process of merging spreg -> pysal. 

This likely only works on *NIX, since I'm using `subprocess`. Also likely only works in `python2` because it's proessing the output of `Popen` as naive strings. 

From Pysal's git root, it works like:
`tools/spregsync path/to/spreg/econometrics path/to/pysal/spreg [--verbose, -v, --dry_run]`. 

This would take the files tracked by git in `pysal/spreg`, and copy the version from `spreg/econometrics` to `pysal/spreg`. Anything not tracked in PySAL that needs to be added from `spreg/econometrics` would need to get added manually .

You can also call
`tools/spregsync --help` to get the builtin help. 

It's using [`click`](https://click.pocoo.org) for the CLI.

### DIFF
--- a/tools/spregsync
+++ b/tools/spregsync
@@ -1,0 +1,47 @@
+#!/usr/bin/python2
+
+import subprocess as sb
+import shutil
+import sys
+
+try:
+    import click
+except ImportError:
+    print('Interface built with Click: \n\tclick.pocoo.org')
+    print('Install click to use (available via pip/conda)')
+    sys.exit()
+
+@click.command()
+@click.option('--dry_run', is_flag=True, help='avoid running copy operation')
+@click.option('--verbose', is_flag=True, help='print files being copied')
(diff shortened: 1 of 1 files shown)


tools/spregsync
```diff
@@ -0,0 +1,47 @@
...
+
+try:
+    import click
+except ImportError:
+    print('Interface built with Click: \n\tclick.pocoo.org')
+    print('Install click to use (available via pip/conda)')
+    sys.exit()
+
+@click.command()
+@click.option('--dry_run', is_flag=True, help='avlid actually executing the copy')
```
ljwolf: looks like a typo

PR #727, opened 2015-12-16
Geodf io

ljwolf:
This includes a number of new pandas tools for shapefile IO. They're loaded in `pysal.contrib.pdio`.

Since these are explicitly paired to our shapefile readers, they only go shapfile-to-dataframe and dataframe-to-shapefile using our IO. I've also got IO-swapped geopandas sitting [here](https://github.com/ljwolf/pysal/tree/geodf/pysal/contrib/geodf), but this (I think) gets us closest to what we want in the smallest amount of code, a raw pandas dataframe, filled with properties & pysal shapes.

### DIFF
--- a/pysal/__init__.py
+++ b/pysal/__init__.py
@@ -84,6 +84,11 @@ import pysal.spreg
 import pysal.examples
 from pysal.network.network import Network, NetworkG, NetworkK, NetworkF
 
+try:
+    import pandas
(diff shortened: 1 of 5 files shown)


ljwolf:
Just to emphasize, these are in *contrib*, and the pandas import in `__init__.py` should correctly catch `ImportErrors`.

PR #732, opened 2016-01-07
Update moran.py with much faster iterations

andrewxhill:
Here is an interesting optimization that @ohasselblad and I have been toying with. Basically, it takes into effect that fact that you are shuffling (at times very large) arrays for every iteration of the loop. When in reality, you should be limiting your necessary shuffle to only the number of permutations or the length of the array, whichever is less.

since it uses the magic of index pointers in numpy, the choice sample will still act across the entire population, while the rids points to that new subsample instead.

### DIFF
--- a/pysal/esda/moran.py
+++ b/pysal/esda/moran.py
@@ -671,15 +671,17 @@ class Moran_Local:
         prange = range(self.permutations)
         k = self.w.max_neighbors + 1
         nn = self.n - 1
-        rids = np.array([np.random.permutation(nn)[0:k] for i in prange])
+        rids = np.random.randint(0, np.minimum(nn, self.permutations), (np.minimum(nn, self.permutations), k))
         ids = np.arange(self.w.n)
         ido = self.w.id_order
         w = [self.w.weights[ido[i]] for i in ids]
         wc = [self.w.cardinalities[ido[i]] for i in ids]
 
         for i in xrange(self.w.n):
-            idsi = ids[ids != i]
-            np.random.shuffle(idsi)
+            idsi = np.random.choice(
+                ids[ids != i], 
+                np.minimum(nn, self.permutations)
+               )
             tmp = z[idsi[rids[:, 0:wc[i]]]]
             lisas[i] = z[i] * (w[i] * tmp).sum(1)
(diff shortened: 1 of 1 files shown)


ljwolf:
Hey, thanks for this! Not sure why travis is erroring out... 

I think we might need to update our contribution directions! Active development occurs against the `dev` branch, and `master` is used for release candidates. 

I've merged your edits into our dev branch [here](https://github.com/ljwolf/pysal/tree/moran-opts), which has run on our [CI](https://travis-ci.org/ljwolf/pysal/builds/100931520).

It seems like there are a few test failures with this change that'll need a closer look, revolving around broadcasting & significance for Moran Rates

andrewxhill:
Ah bummer, I got excited and did it in haste with total disregard to your contributions.md. Sorry abou that :( 

If easier, I don't mind if you want to boot it out and I can do it again the right way. 

It will be interesting to take a look at. It takes advantage of that neat little cap on permutations. I believe it is still correct, but you may find an issue.

lixun910:
Just take a look at this fix, one issue: the purpose of numpy.permutation in Moran's I calculation is for generating random neighbors:
rids = np.array([np.random.permutation(nn)[0:k] for i in prange])

However, by using np.random.randint, you can't guarantee that there is no duplicates in each set of random neighbors, since randint will draw same number more than one time, especially given a small set of numbers.
rids = np.random.randint(0, np.minimum(nn, self.permutations), (np.minimum(nn, self.permutations), k))

So, the logic here is not correct, and cause the check failure.

andrewxhill:
Ah okay, nice. The key change in the rids constructor here is to limit the identifiers to the min of permutations or nn. So it should actually be,

rids = np.array([np.random.permutation(np.minimum(nn, self.permutations))[0:k] for i in prange])

ljwolf:
Ok, I'm returning to this post release. Time permitting, I'll open a new PR with @andrewxhill's changes, rebased ontop of PySAL 1.11. 

Don't want this to get stale.

ljwolf:
Ah, I didn't realize it's already been rebased. Cool! I'll take a look at the new test failures and see what we can do.

ljwolf:
Rebased & submitted to pysal/dev via #744, closing this.

PR #735, opened 2016-01-27
Dev in sync with master for 1.11

sjsrey:


ljwolf:
Uh oh, this made dev 1 in front of, 1 behind master... I think we need to do this from pysal/master, rather than from serge/dev

sjsrey:
I will rectify.

PR #743, opened 2016-02-15
[REBASED]: Update moran.py with much faster iterations

ljwolf:
From @andrewxhill, rebasing #732 onto PySAL 1.11. 

The rebase still appears to be failing/erroring out on some [moran-related tests](https://travis-ci.org/ljwolf/pysal/jobs/109445428), but those could just be plumbing issues. 

> Here is an interesting optimization that @ohasselblad and I have been toying with. Basically, it takes into effect that fact that you are shuffling (at times very large) arrays for every iteration of the loop. When in reality, you should be limiting your necessary shuffle to only the number of permutations or the length of the array, whichever is less.
> 
> since it uses the magic of index pointers in numpy, the choice sample will still act across the entire population, while the rids points to that new subsample instead.

### DIFF
--- a/pysal/esda/moran.py
+++ b/pysal/esda/moran.py
@@ -671,15 +671,17 @@ class Moran_Local:
         prange = range(self.permutations)
         k = self.w.max_neighbors + 1
         nn = self.n - 1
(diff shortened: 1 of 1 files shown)


ljwolf:
Ugh, wrong target.

PR #744, opened 2016-02-15
[REBASED] Update moran.py with much faster iterations

ljwolf:
From @andrewxhill, rebasing #732 onto PySAL 1.11. 

The rebase still appears to be failing/erroring out on some [moran-related tests](https://travis-ci.org/ljwolf/pysal/jobs/109445428), but those could just be plumbing issues. 

> Here is an interesting optimization that @ohasselblad and I have been toying with. Basically, it takes into effect that fact that you are shuffling (at times very large) arrays for every iteration of the loop. When in reality, you should be limiting your necessary shuffle to only the number of permutations or the length of the array, whichever is less.
> 
> since it uses the magic of index pointers in numpy, the choice sample will still act across the entire population, while the rids points to that new subsample instead.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,6 +1,6 @@
 v<1.11.0>, 2016-01-27
 
-GitHub stats for 2015/01/31 - 2015/07/29
+GitHub stats for 2015/07/29 - 2016/01/27
 
 These lists are automatically generated, and may be incomplete or contain duplicates.
 
--- a/pysal/esda/moran.py
+++ b/pysal/esda/moran.py
@@ -671,15 +671,17 @@ class Moran_Local:
         prange = range(self.permutations)
         k = self.w.max_neighbors + 1
         nn = self.n - 1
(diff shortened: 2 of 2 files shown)


ljwolf:
So, I'd recommend closing, since it seems sampling without replacement is just as slow as permuting the entire matrix, unless someone has the time to do a more thorough comparison on that `choice` vs. `shuffle` performance.

andrewxhill:
@sjsrey, i believe xun's comment was about 1 line that isn't actually the heart of the enhancement, I fixed that in a commit following,

https://github.com/pysal/pysal/pull/732#issuecomment-170037766

There would be no improvement for small datasets (<1000) but it helps cap out the duration for very large datasets. The core is this,

"Basically, it takes into effect that fact that you are shuffling (at times very large) arrays for every iteration of the loop. When in reality, you should be limiting your necessary shuffle to only the number of permutations or the length of the array, whichever is less."

https://github.com/andrewxhill/pysal/commit/bcb169c7a514095575418948dd569729520c0039#diff-505086ccec22587314a0b60c73239426R682

PR #753, opened 2016-03-01
Fix the network ring bug

### LINKED ISSUE #655: network ring bug
In the case of a ring road with no intersections, the Network builder fails.  The traceback is not informative to the user.  This is a rare case, so is a low priority.

code

ring_road

chiragvartak:
Fixes #655, and some other minor changes are made:
(1) Added a parameter 'own_edges' to the Network class constructor
(2) Modified the code to work with shapefiles that have shapes with
multiple parts
(3) Added an example to pysal/examples


ljwolf:
Thanks for the patch, @chiragvartak! Unfortunately, some of the CI tests are failing, and I think we need to resolve those before there's any merge. To fix them, it looks like you could just change the test answer, since we're changing the way the graph gets simplified. But, we'd need to check that the results of those algorithms are accurate, given the change in simplification. 

But, I think #655 also needs to be answered before we merge this :disappointed:

ljwolf:
Also, just to be sure, to check out the tests yourself, check the "details" link below. I've circled it in this image. 

2016-03-01-182655_905x515_scrot

sjsrey:
Closing. Will pick back up in #655

PR #755, opened 2016-03-07
add speedup of conditional randomization

ohasselblad:


ljwolf:
Hey, thanks for this @ohasselblad! We're having some trouble with our CI.. looks like something to do with [conda/#208](https://github.com/conda/conda/issues/2028). Nothing on your end :smiley: 

To the substance, the conditional permutation routine really needs two guarantees on its sampling for "full soundness"

1. No self-sampling
2. No duplicates in resluting sample (i.e. sampling w/o replacement). 

I see that [this removes self-samples](https://github.com/pysal/pysal/pull/755/files#diff-505086ccec22587314a0b60c73239426R683), so 1 is guaranteed. And, I recognize that the probability of `np.random.choice` returning a duplicate entry for large data approaches zero. But sampling with replacement (the default for `np.random.choice` and what's in this PR) isn't "correct" in the general case, even though it gives us a *huge* speedup (~5 orders of magnitude, as I show below)

So, to satisfy point 2, we need `np.random.choice(..., replace=False)`, or a way to guarantee that the resulting sample has no duplicates and no self-sample. Unfortunately, in the source cython, `permutation(...)[:n]` [is actually what gets called in the end](https://github.com/numpy/numpy/blob/master/numpy/random/mtrand/mtrand.pyx#L1449) if you call `choice(...,size=n, replace=False)`. This means that sampling without replacement and permuting the whole matrix are hitting *the exact same code points* in numpy, so we'd expect their timings to be nearly equivalent in general: 
2016-03-07-114615_881x307_scrot
You can rerun these timings but, in general, sampling w/o replacement & permutation will be equivalent in the long run because they're running the same cython, but sampling w/ replacement will be very fast in comparison.

ljwolf:
What I think might help everyone here would be to do exactly what y'all've submitted, but then check the sample for validity & fix duplicates. This would, in theory, be nearly as fast in the small data case and, in the big data case, could still retain the benefits of sampling w/o replacement in large datasets, as long as `np.unique` is fast enough.

So, once we fix the stuff with the CI, I'm suggesting something like: 
```
idsi = np.random.choice(ids[ids != i], np.minimum(nn, self.permutations))
uvals, uidxs = np.unique(idsi, return_index= True)
if uidxs.shape != idsi.shape:
    newpop = np.setdiff1d(ids[ids != i], uvals) #members of pop not in {i, uvals}
    idsi[~uidxs] = np.random.choice(newpop, len(uidxs), replace=False) #expensive sample the smaller set
```
Thus, the sampling w/o replacement only hits if we get a bad sample. 

If the speed of `choice + unique` is faster than `choice(..., replace=True)`, this could work.

ljwolf:
Preliminary testing shows this `sample + validate` strategy is ~3 orders of magnitude faster than simple samping w/o replacement when there are no collisions, on the order of 300 milliseconds for the same data shown above.

ohasselblad:
Oof, I accidentally PR'd to your repo instead of my fork! Sorry, closing :) And taking a look at your comments, @ljwolf.

ljwolf:
Ah, okay! 

Well, let me know if there's anything I can do to help. 

By the way, @sjsrey just fixed our CI, so if you rebase to our current dev, the tests should work.

ohasselblad:
@ljwolf, nice piece of code -- we'll use that one in our testing :+1:

ohasselblad:
@ljwolf, testing that out, I'm seeing times when `len(uidxs)` is larger than `len(newpop)`, which is not allowed for `np.random.choice` when `replace=False`.

Should we just be looking at the full set of `newpop` but randomly sampled, like this:
```
idsi[~uidxs] = np.random.choice(newpop, len(newpop), replace=False)
```

ljwolf:
Ah,  I see. Sampling `len(newpop)` from `newpop` will permute the remaining table, I think. Don't want that... 

I'd hoped to sample only the number of invalid (duplicate) members of the original sraw. So, that (I think)  is `len(idsi[~uidxs])` or `len(idsi) - len(uidxs)` instead of just `len(uidxs)`. My bad!  

In the end, if `len(isdsi) == np.minimum(nn, self.permutations)` with no repeats or self-draw, the sample should be okay.

PR #756, opened 2016-03-07
B726

sjsrey:
Some fine tuning in spreg/tests that will have to go back upstream to spreg, but travis passes now with unpinned scipy and numpy to development can move forward.


ljwolf:
:tada: :balloon: :tada: will merge when passing. 

We should still look into using CircleCI for the private repositories. I can get circleCI running on my private spreg fork, but cannot test out how to set it up for `pysal/spreg`, since I'm not an admin.

pedrovma:
Brilliant!

sjsrey:
For future reference, when we are testing allclose and one of the parameters in the array is close to (or equal to) 0, just using rtol will lead to problems, since rtol*v will be 0. In that case using atol will be the way to go.

pedrovma:
:+1:

PR #757, opened 2016-03-10
Esda shuffle speedup

andy-esch:
Ref https://github.com/pysal/pysal/pull/755

/cc @ljwolf

### DIFF
--- a/pysal/esda/moran.py
+++ b/pysal/esda/moran.py
@@ -683,12 +683,14 @@ class Moran_Local:
             # orig source
             # idsi = ids[ids != i]
             # np.random.shuffle(idsi)
-            idsi = np.random.choice(
-              ids[ids != i],
-              np.minimum(nn, self.permutations)
-            )
-            tmp = z[idsi[rids[:, 0:wc[i]]]]
-            lisas[i] = z[i] * (w[i] * tmp).sum(1)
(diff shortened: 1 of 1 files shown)


sjsrey:
Revisit at Scipy2016

ljwolf:
Okay, so after some inspection, I'm concerned that something's up. Namely, when we use this patch over more data than the unittest, the expected change in the `p_z_sim` tend to be negative, so this technique, on average, generates smaller values. I'm not sure why this is the case from the code, so we're going to check them against the p-values in R for the same datasets.  

Thanks for bearing with us... this has taken a long time to review because these stats are pretty core to users & they also receive a lot of academic attention. We want to merge this, but have to make sure the patch isn't somehow skewing the `p_z_sim` estimates in a predictable way.

ljwolf:
As a note to a contributor/maintainer who will pick this up:
- check pvalues over common datasets in:
    - [ ] `R` using `spdep`
    - [ ] OpenGeoDa
- ensure `p_z_sim` or equivalent statistics are within reasonable statistical distance to other packages.

andy-esch:
Not a priority?

ljwolf:
We need to direct it to [pysal/esda](https://github.com/pysal/esda) as the `dev` branch has been deleted. This closed by default when I deleted the unused `dev` branch.

Sorry about that...

Issue #767, opened 2016-03-20
ZeroDivisonError when calculating certain centroids


stefanborchardt:
I can calculate all centroids of a shapefile with shapely 1.5.13, but PySAL 1.11 throws ZeroDivisonErrors. Geos version is 2.0.2, I think. Python 2.7.11/ Anaconda 2.5.0 64bit.

Shapefile problem polygons: 
[zerodivision_shpfile.zip](https://github.com/pysal/pysal/files/181273/zerodivision_shpfile.zip)
Pickle of the pysal (and shapely) polygons that cause trouble:
[centroid_divzero.zip](https://github.com/pysal/pysal/files/181247/centroid_divzero.zip)
Full data set https://data.sfgov.org/Energy-and-Environment/San-Francisco-Urban-Tree-Canopy/w9tk-3w8c

The error occurred when I used pysal.min_threshold_dist_from_shapefile(). It would be great if such a rather high level function would handle things more gracefully.

kritinsai:
Could you list the modules you imported when you created the pickle, cause i am not able to load it.

kritinsai:
@stefanborchardt Could you list the modules you imported when you created the pickle, cause i am not able to load it.

stefanborchardt:
@kritinsai Sorry.
`import pysal
import shapely.geometry as sg
import shapely.ops as so
import fiona
import pickle`

kritinsai:
@stefanborchardt Is 75461932.639456302 a reasonable value for the minimum_threshold_distance?

stefanborchardt:
@kritinsai, I have to admit I'm new to spatial analysis. My calculation for the part of data that did not throw an error was 934258.692654. Coordinates are in a state plane with a bounding box of about [5979385,2085840,6024664,2123823]. I would expect a minimum threshold to be lower than the width of 40000 - the number says to what distance polygons should be regarded as belonging together, right? Maybe 40?

Makosak:
I had a similar issue last year; there may be an error with the shapefile you're using. I was using a standard NYC shapefile downloaded from a gov org, and there were digitizing issues that resulted in this error being thrown when I tried to calculate a K-nearest neighbors function that required centroids. The shapefile reports a bounding box that does not contain all the bounding boxes for the individual shapes.

Try validating the geometry; following comments I received back then:
QGIS -> Vector -> Geometry Tools -> Check Geometry Validity

ljwolf:
This may be fixed by #768

PR #768, opened 2016-03-20
Modified the way area of a ring is calculated to allow for more precision

kritinsai:
#767 : The problem is when the area of a ring is calculated: Precision is lost somehow and area becomes 0 for very small rings. This problem does not occur when the area of a Polygon is calculated which contains this single ring (the method used is different and retains precision). Hence the same method can be used to calculate the area of a ring.

### DIFF
--- a/pysal/cg/shapes.py
+++ b/pysal/cg/shapes.py
@@ -1280,7 +1280,7 @@ class Ring(object):
                 A += (x[i] + x[i + 1]) * \
                     (y[i] - y[i + 1])
             A = A * 0.5
-            self._area = A
+            self._area = -A
         return self._area
 
     @property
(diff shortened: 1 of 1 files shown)


ljwolf:
In the [testing] (https://travis-ci.org/pysal/pysal/jobs/117270136) this change generates the negative of some test shape areas, like [l424](https://travis-ci.org/pysal/pysal/jobs/117270136#L424). It might be good to review the tests and see where that inversion is happening. Off the top of my head, I don't see why that would occur.

kritinsai:
The sign of the area which my modified method returns is opposite to the expected sign. In the centroid method of the ring, Ring.signed_area is used, and it is also the only place where it is used, so I modified the centroid method to use only the vertices of the ring, this also handles the case when the area of the ring is quite small that python considers it to be zero. -- turns out this leads to a lot of difference  in the calculated values in all the automated tests. @ljwolf @sjsrey Any suggestions?

ljwolf:
Yeah, it looks like it changes the centroids of all shapes, which changes any result that depends on it, like Kernel, HAC, and threshold distance. \

The method you've used looks much more robust to float point accumulation errors... but since changing potential floating point accumulation errors in computing the centroid will hit a ton of tests, we'll definitely need to review this further. 

I also wonder if it's possible to just handle the original zerodiv issue without totally changing the results of the centroid computation. I've just checked our centroids against a few computed in Shapely and the results are within float error.

ljwolf:
@sjsrey thoughts? This is pretty far-reaching.

kritinsai:
The issue with the inverse sign could be because of the way the calculations are done in Ring.centroid and my old Ring.signed_area.

sjsrey:
Will revisit at SciPy2016 sprint

ljwolf:
The math looks correct, and I've restarted the CI. This should get merged once travis reruns?

Issue #771, opened 2016-03-23
KDtree type mismatch in knnW


ljwolf:
From pysal-dev:

> When I am running the following code for weight generation
> ```python
> import pysal
> import numpy as np
> x,y = np.indices((5,5))
> x.shape = y.shape = (25,1)
> data = np.hstack([x,y])
> wknn3 = pysal.knnW(data, k=3)
> wknn3.neighbors[0]
> wknn.s0
> ```
> then I am getting the following error:
> ```python
Traceback (most recent call last):
  File "C:\Python27\FOSS4G NA 2015 - Spatial Data Analysis in Python\PySAL_Workshop\mine\knnweight.py", line 9, in <module>
    wknn3 = pysal.knnW(data, k = 3)
  File "C:\Python27\FOSS4G NA 2015 - Spatial Data Analysis in Python\PySAL_Workshop\mine\pysal\weights\Distance.py", line 86, in knnW
    nnq = kdtree.query(data, k=k+1, p=p)
 AttributeError: 'numpy.ndarray' object has no attribute 'query'
```
> Why? Please help!

ljwolf:
The error actually occurs right at the `knnW` weights constructor.
It looks like commit [a38c7d2c](https://github.com/pysal/pysal/commit/a38c7d2cd3df3bdf6d505e36c707d0fafdc16ba0#diff-bec4c529c8f5a2e5515050c698b4c902R85) removed dispatch on type of input for the knnW constructor. 

It used to accept numpy arrays and silently build the KDTree behind the scenes. Now, it will fail immediately if the user passes a previously-valid input.

Potential issues with this practice. 

1. We never issued any deprecation warnings on this change. 
2. This thumps a [2015 FOSS4G workshop](https://2015.foss4g-na.org/session/hands-introduction-spatial-data-analysis-python)'s code
3. Previously, proccessing ndarrays `reasonably` as the source for what the user *intends* is a KDTree obeys [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle), but definitely includes unnecessary chrome around the core algorithm. (this is where I like multiple dispatch :)

But, I see nothing wrong with @jlaura's commit itself. It's just an API change that was not smoothed by a deprecation warning. 

I'd mark this as notabug+wontfix, but perspective is needed.

schmidtc:
The OP's code was copy/pasted from the user guide: http://pysal.readthedocs.org/en/latest/users/tutorials/weights.html#k-nearest-neighbor-weights

[...]

PR #773, opened 2016-03-23
add isKDTree typecomparison to handle divergent cKDTree and KDTree types

ljwolf:
This addresses #771 

I have reverted the `kdtree`-only behavior in `knnW` introduced earlier, for the sake of where we've agreed weights module should go: constructors handling *arbitrary intelligble input*. 

I've also addressed the bug where using a `KDTree` constructed using `pysal.cg.KDTree` or `scipy.spatial.cKDTree` will cause incorrect type dispatching in the distance weights constructors. 

I suggest that we examine adding deprecation warnings to type-specific functions like `knnW_from_array` & `queen_from_shapefile` to notify users of this planned change.

### DIFF
--- a/pysal/weights/Distance.py
+++ b/pysal/weights/Distance.py
@@ -11,11 +11,11 @@ from pysal.common import KDTree
 from pysal.weights import W
 import scipy.stats
 import numpy as np
+from util import isKDTree
 
 __all__ = ["knnW", "Kernel", "DistanceBand"]
 
-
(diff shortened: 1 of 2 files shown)


ljwolf:
by reverted, I don't mean I've actually reverted @jlaura's commit, just to be clear :smile:

ljwolf:
Merging these changes with @dfolch's changes at aag

Issue #780, opened 2016-04-01
Doctests failing on travis


sjsrey:
[Turning the doctests back on](https://travis-ci.org/pysal/pysal/builds/119891949) results in failures on travis.  Should https://github.com/pysal/pysal/commit/5e54df67cce183685fe9c492c96ac36a5a093fce be reverted and moved to a branch  off of dev where it can be fixed?

PR #781, opened 2016-04-01
Dev

sjsrey:
Turning doctests back on so dev can pass. Will start a branch to fix doctests ala #780


ljwolf:
Looks good. Someone should start a feature branch to handle doctesting going back on?

PR #793, opened 2016-04-12
Getting weights doctests to pass

sjsrey:
These now pass locally. doctests are still off on travis and will stay that way while other modules get fixed locally.


ljwolf:
sweet! this is great! the issue with `lag_categorical` was with seed setting in the docstring?

sjsrey:
@ljwolf it was the seed, but also the fragile nature of the format that sphinx seems to accept. Hence putting in explicit tests of equality for arrays rather than testing on the formatted output.

PR #798, opened 2016-04-22
fix for css problem on rtd #790

### LINKED ISSUE #790: readthedocs css missing in action?
After the 1.11.1 [release](https://github.com/pysal/pysal/commit/29f85b8c74053d40ea10533094a7ccfce00d8f5f) it appears the automatic build of our docs on [readthedocs](https://pysal.readthedocs.org/en/latest/) is not rendering properly.

When the docs are built locally, there are no errors, but the hosted version does some stuff server side that we can't/don't test locally.

sjsrey:
The justification for this PR is: This appears to fix #790


ljwolf:
I am building locally now. Is there a way to do a test build actually on RTD?

ljwolf:
Locally, the theme swap looks like it's working.

PR #802, opened 2016-05-03
fixed issues with model handler failing to correctly discover models

ljwolf:
The justification for this PR is: 

The `handler` contrib module, which is a Patsy/Pandas wrapper for spreg, had a pretty severe regression of unknown origin. This caused `registry.py` to fail to discover all models in `pysal.spreg`, and resulted in complete loss of functionality for the handler.

This was remedied by restricting the handler to only userclasses, changing the way arguments were parsed, and adding some name parsing to the patsy/pandas preparation step.

As a side effect, this PR also fixes the issue noted in the original specification, that additional unnamed positional arguments would pass silently through when the user supplied a formula. Now, this will not be the case. 

In addition, tests were adapted to make sure the handler is now tested for all user classes in spreg. But, since this is in contrib, the handler won't get tested unless explicitly checked.


ljwolf:
The notebook is screwed up. Let me patch it before merging, please!

ljwolf:
Ok, the notebook should be fixed.

PR #819, opened 2016-06-08
add LIMAs

weikang9009:
### Add 4 concordance statistics proposed by Rey (2016)
1. Local Tau
2. Neighbor set LIMA
3. Neighborhood set LIMA
4. Inter and Intraregional decomposition of Tau


sjsrey:
We should move this to a different branch as there is some refinement and additional testing that need to be fleshed out before this goes into pysal:dev.

Please create a branch "lima" in your repos and you and I can collaborate on that. Once you do, I will close this pr and we will pick up on your branch.

Issue #826, opened 2016-06-28
Spatial Filtering grid definition


ljwolf:
The spatial filtering grid definition returns `(n+1)**2` grid cells rather than `n` grid cells sometimes because of striding. This is another one of those tricky `(int/int)` bugs that's lurking from python 2.  For example, the call in the unittest is:

```python
bbox = [(0,0), (45,45)]
pysal.esda.smoothing.Spatial_Filtering(bbox, data, e,b, x_grid=2, y_grid=2, r=30)
```
which should divide the bounding box provided into 2x2 grid.
Instead, the unittest thinks the answer is a 3x3 grid.

This is because the number of divisions doesn't evenly divide into the range of the bounding box, **and** the bounding box is integer. [The call is effectively](https://github.com/pysal/pysal/blame/master/pysal/esda/smoothing.py#L1197):

```python
np.mgrid[min_y:max_y:(range_y / y_grid)]
```
In the unittest, `(range_y/y_grid)` is integer division, since the bounding box is integers and the `y_grid` is integer. This means we get 22, rather than 22.5, and the grid produced is:
```python
>>> np.mgrid[0:45:(45 / 2)]
np.array[(0, 22, 44)]
```
rather than what we want:
```
>>> np.linspace(0, 45, num=2, endpoint=False)
np.array([0, 22.5])
```

This means the test is incorrect. I think this was first noticed in the lead up to the python 3 conversion sprint at scipy last year, but the test should've been rewritten, rather than using floor division. Should be an easy fix, but it'll mean rewriting the test.

ljwolf:
This should be fixed in #827:

```python
import pysal as ps
bbox = [(0,0), (45,45)]
df = ps.open(ps.examples.get_path('sids2.dbf')).to_df()
e,b = df.SID74, df.BIR74
data = df.geometry.apply(lambda x: x.centroid).tolist()
sm = ps.esda.smoothing.Spatial_Filtering(bbox, data, e, b, x_grid=2, y_grid=2, r=30)
assert len(sm) == 4
```

Issue #825, opened 2016-06-26
Headbanging Median Rate ignores edge correction


ljwolf:
This bug took a long time to figure out what was going on, but I think I have a diagnosis. A patch is coming in my [GSOC feature branch](https://github.com/ljwolf/pysal/tree/pdio) 

`Headbanging_Median_Rate` ignores edge correction information in `Headbanging_Triples` instances: it looks for [`exta`, not `extra`](https://github.com/pysal/pysal/blob/26c23b0bd86cfa1886bb5d722ab02142d902b090/pysal/esda/smoothing.py#L1496)

So, if anyone does the following:
```python
hbt = ps.esda.smoothing.Headbanging_Triples(data, w, edgecor=True)
hbr = ps.esda.smoothing.Headbanging_Median_Rate(e, b, hbt)
```
the resulting headbanging median rate will silently ignore the edge correction in the headbanging triples argument. 

When I fixed this typo, there were a few unaddressed bugs in the `extra`-related paths of the `Headbanging_Median_Rate` helper functions `__get_screens` and `__get_median_from_screens`. 

As far as I can tell,  [L1689](https://github.com/pysal/pysal/blob/26c23b0bd86cfa1886bb5d722ab02142d902b090/pysal/esda/smoothing.py#L1513) & [L1694](https://github.com/pysal/pysal/blob/26c23b0bd86cfa1886bb5d722ab02142d902b090/pysal/esda/smoothing.py#L1518) should return `r[id]` (like the rest of the return statements) instead of `r`.

ljwolf:
This should be resolved in #827.

Issue #824, opened 2016-06-25
Direct Age Standardization fails for empty regions


ljwolf:
I think there's a minor bug in `pysal.esda.smoothing.direct_age_standardization`:

```python
for i in range(len(adjusted_r)):
    if adjusted_r[i] == 0:
        upper = .5 * scipy.stats.chi2(1-0.5 * alpha)
        lower = 0
    else:
        lower = scipy.stats.gamma.ppf(.05 * alpha, g_a[i],. scale=g_b[i])
        upper = scipy.stats.gamma.ppf(1 - .05*alpha, g_a_[k], scale=g_b_k[i]
```
I think that should be `scipy.stats.chi2.ppf`, as is the case for the gamma function. As it stands, when `adjusted_r[i] == 0`, we get a `TypeError`, since calling `chi2(parameters)` directly creates a frozen random variable, which can't be multiplied. 

The traceback that occurs is: 

```python
/home/ljw/dev/pysal/pysal/esda/smoothing.pyc in __init__(self, e, b, w, s, alpha)
    978         e_n = np.array(e_n).reshape((1, t), order='F')[0]
    979         b_n = np.array(b_n).reshape((1, t), order='F')[0]
--> 980         r = direct_age_standardization(e_n, b_n, s, w.n, alpha=alpha)
    981         self.r = np.array([i[0] for i in r])
    982         w.transform = 'o'

/home/ljw/dev/pysal/pysal/esda/smoothing.pyc in direct_age_standardization(e, b, s, n, alpha)
    283     for i in range(len(adjusted_r)):
    284         if adjusted_r[i] == 0:
--> 285             upper = 0.5 * chi2(1 - 0.5 * alpha)
    286             lower = 0.0
    287         else:

TypeError: unsupported operand type(s) for *: 'float' and 'rv_frozen'
```

Someone with a knowledge of smoothing should probably review.

ljwolf:
should be resolved in #827

PR #827, opened 2016-06-30
ESDA Tabular Functions

ljwolf:
Hello!

This is the work I've done to get tabular functions into esda. I'd hoped to at least submit this before going off into the rest of the project on the computational geometry stuff. Like before, [here's the overview notebook. ](http://nbviewer.jupyter.org/github/ljwolf/pysal/blob/pdio/ESDA%20Overview.ipynb)

 A few things:

1. separating the `by_col` methods out from the classes proved to be too difficult. In some cases, there would have been special `by_col` functions with dispatch on the input type, like for `Age_Adjusted_Smoother`. So, I just added them as class methods & added unittesting by class. 
2. This also addresses the bugs I encountered in smoothing, #824 #825 #826.


ljwolf:
sorry, looks like there's some cruft from my pivot to the computational geometry stuff that snuck in, like `ogc_synthesis` and some undesirable changes to `travis.yml`. 

will need to remove it before merging.

ljwolf:
OK, I think this is ready for review.

PR #829, opened 2016-07-08
numba autojit _fisher_jenks_means if numba is available

mlyons-tcc:
fixes 828. gets over 100x performance improvement for _fisher_jenks_means if numba is installed.

### DIFF
--- a/pysal/esda/mapclassify.py
+++ b/pysal/esda/mapclassify.py
@@ -21,6 +21,11 @@ import copy
 import sys
 from scipy.cluster.vq import kmeans as KMEANS
 from warnings import warn as Warn
+try:
+    from numba import autojit
+except ImportError:
+    def autojit(func):
+        return func
 
 def headTail_breaks(values, cuts):
     """
@@ -313,6 +318,7 @@ def natural_breaks(values, k=5):
     return (sids, class_ids, fit, cuts)
 
 
+@autojit
 def _fisher_jenks_means(values, classes=5, sort=True):
     """
     Jenks Optimal (Natural Breaks) algorithm implemented in Python.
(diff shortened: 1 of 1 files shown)


sjsrey:
Thanks @mlyons-tcc - this looks really promising.

Can you add a test suite for this enhancement, in [esda/tests/test_mapclassify.py](https://github.com/mlyons-tcc/pysal/blob/2186705e018a7e35f790f6b5e5a6a4de48a3b6e2/pysal/esda/tests/test_mapclassify.py) ?

ljwolf:
Yeah, this looks real cool @mlyons-tcc, thanks!

I've been looking at doing the same thing in `cg`, but haven't seen this kind of speedup. Jenks is a great target for this. 

Locally, I go from ~10 seconds/run on the `NAT` `HR90` attribute to ~80 ms.

mlyons-tcc:
Thanks. I tested out a Cython implementation that performs faster. Numba was on the same order of magnitude though, and I really like the `autojit` decorator. It's a shame that you have to jump through so many hoops to get Numba installed if you aren't using Conda. I'm not good enough at Cython to get [Pure Python Mode](http://docs.cython.org/src/tutorial/pure.html) working.

@sjsrey, there is already a test: TestFisherJenks that tests the algorithm.

If I understood correctly PySAL requires not depending on external packages other than numpy and scipy so I didn't force a numba dependency. I'm not experienced in using travis.yml scripts so I don't know the proper way to run tests with/without numba. It also seems weird to run the whole test suite twice for this one function with/without numba optimization. 

When numba is installed, it just does an `autojit` on the one function. In my opinion, it's highly unlikely the function will not work properly when bypassing autojit if the autojit version is tested. However, it's possible for the autojit to break and not work despite the algo passing its test when autojit is bypassed. Because of that, I would recommend that I just add `numba` to the `conda install` section of the `.travis.yml` file.

Please advise.

sjsrey:
I like this approach, but am thinking about two things: [1] whether we should use the same strategy more broadly throughout PySAL (where numba would pay off) and thus centralize the numba check a la [this approach](https://github.com/soft-matter/trackpy/blob/master/trackpy/try_numba.py); and [2] how to modify our tests for the numba present and numba absent cases.

ljwolf:
As @mlyons-tcc notes, it'd be real weird if `autojit` changed the results of the computation, but I understand the desire to test it just in case. 

I'm working on setting up a [test matrix](https://travis-ci.org/ljwolf/pysal/builds/143763824) to address this concern. Since we're going to need to do this anyway empower new soft dependencies, it's needed regardless of this contribution.

If I submit the test matrix PR, can we merge this?

sjsrey:
+1

PR #830, opened 2016-07-12
[REBASED] PySAL+ optional testing matrix

ljwolf:
This sets up a multi-environment test matrix for PySAL using an environment variable, `PYSAL_PLUS` & is the remaining blocker for #829 

If `PYSAL_PLUS`, we include pandas, numba, numexpr, shapely, and patsy in the testing environment. This way, if we have numba versions of computational functions, tabular methods, etc, these get tested. 

In theory, these should also be added to the `setup.py` `extras_require` flag, if we want to appropriately incorporate this metadata by module. 

It makes sense to make the `PYSAL_PLUS` environment as open as possible, as long as things were correctly wrapped in try/except or `common.requires`. <s> I will write that up in a wiki page, too. </s> I've written [this notebook](http://nbviewer.jupyter.org/gist/ljwolf/b9817ce4590994f20a77f783a9bff3f4) documenting things I've seen & tried.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ before_install:
   - ./miniconda.sh -b -p ./miniconda
   - export PATH=`pwd`/miniconda/bin:$PATH
   - conda update --yes conda
+  - conda config --add channels conda-forge
   - conda create -y -q -n test-env python=$TRAVIS_PYTHON_VERSION
   - source activate test-env
   - if [[ $TRAVIS_PYTHON_VERSION == 3* ]]; then 2to3 -nw pysal/ > /dev/null; fi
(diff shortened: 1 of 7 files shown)


ljwolf:
adding tentative label in case anyone wants to call out extra packages to add to `PYSAL_PLUS`.

ljwolf:
Just realized, visualization packages might be good to add here. At minimum, matplotlib & seaborn.

Makosak:
Matplotlib templates in the new version includes a seaborn styling... Have
to look at that further, but could help

ljwolf:
Rebased into #833.

PR #846, opened 2016-07-20
DistanceBand speed ups

TaylorOshan:
This PR introduces two changes to the DistanceBand class in the weights module. The first change is to the way the neighbors and weights dictionaries are created when constructing the the DistanceBand W object. Instead of using a loop, a pairwise distance matrix is used to create a sparse matrix using ``WSP()`` and then converted to a regular W using ``WSP2W()``. Finally, the neighbor and weight dictionaries from this regular W can be used to instantiate a DistanceBand class W. This provides a speed-up, especially as the number of spatial units increases. 

The second significant change is the addition of an option to build the DistanceBand weight using a dense distance matrix function rather than solely the option to use the sparse distance matrix method from the kdtree. If the user is has a relatively high threshold set, the W will not be that sparse, so using dense operators will result in significant speed-ups. This option can be enabled by changing a new Boolean parameter ``build_sp`` to ``False``. The default is set to ``True`` which maintains the behavior of the DistanceBand class as it was previously. 

Tagging @sjsrey , @carsonfarmer , and @darribas  as this is somewhat related to GSOC goals and tagging @dfolch who was helping with this at Scipy. Also tagging @ljwolf because there may be merge conflicts previously merged PR's from his GSOC work.

### DIFF
--- a/pysal/weights/tests/test_Distance.py
+++ b/pysal/weights/tests/test_Distance.py
@@ -134,7 +134,7 @@ class TestDistanceWeights(unittest.TestCase):
         kd = pysal.cg.kdtree.KDTree(pts, distance_metric='Arc',
                                     radius=pysal.cg.sphere.RADIUS_EARTH_KM)
         w = pysal.DistanceBand(kd, full.max(), binary=False, alpha=1.0)
-        np.testing.assert_allclose(w.sparse.todense(), full)
+        self.assertTrue((w.sparse.todense() == full).all())
 
     def test_DistanceBand_dense(self):
         """ see issue #126 """
--- a/pysal/weights/tests/test_Distance.py
+++ b/pysal/weights/tests/test_Distance.py
@@ -2,6 +2,7 @@ import os
 import unittest
 import pysal
 import numpy as np
+import sys
 
 
 class TestDistanceWeights(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


TaylorOshan:
Not sure if I am missing something, but as far as I can tell, I don't permission to assign or label, so I haven't done so.

pysal/weights/tests/test_Distance.py
```diff
@@ -2,7 +2,7 @@
 import unittest
 import pysal
 import numpy as np
-
+import sys
```
ljwolf: Why is `sys` needed now?

ljwolf:
So, the merge conflicts on this are just in Distance & its unittest. And, the ones in distance are straightforward, since I didn't touch the actual distance matrix logic, which is what this PR enhances. 

It might be easier to just yank the changes to the unittest into a local copy of dev. Let me keep looking at this and I'll update.

ljwolf:
yeah.... I resolved the conflicts, but I'm not sure how to get this around the horn. It's >100 commits if I send it to you via PR.

What if we set aside a time to `git pull --rebase upstream dev` on this, then walk through the merge conflicts together? They're relatively straightforward I think, but it'll help us figure out which goes where *(Also secretly a good reason to try out [madeye.io](https://madeye.io))*

TaylorOshan:
Sounds good, lets set up a time to chat and go through it.

pysal/weights/tests/test_Distance.py
TaylorOshan: Its not, could be removed. I think I was just using to for some local testing.

ljwolf:
This is superseded in #850.

PR #847, opened 2016-07-20
spatial interaction weights 

TaylorOshan:
This PR add several functions for building spatial weights for spatial interaction data. The first type of SI weight, ``ODW()`` creates contiguity weights where non-zero entries represent when a spatial unit is a neighbor to either the origin or the destination. The second type of SI weight, ``netW()`` creates a variety of contiguity weights where spatial units are flows or edges and contiguous is defined as sharing a particular type  of node. Finally, ``vecW()`` can be used to build distance-based weights for a 4-dimensional vector (origin x, origin y, dest x and dest y). 

Tagging @sjsrey @carsonfarmer @darribas as this work is related to GSOC efforts, though perhaps there are others who might be well equipped to evaluate this addition?


sjsrey:
I think this will be really useful. 

It is possible that this functionality might be used to address something @lanselin asked about a while back, and that was related to rectangular weights matrices (i.e., where the origin and destination sets were not necessarily identical.)

TaylorOshan:
It works with different sets of origins and destinations, but I think you always get a square weights matrix. The final shape of the full weights matrix is ``(x,x)`` where ``x = n**2 * m**2`` and n and m are the number of origins and destinations, respectively.

darribas:
Wouldn't `x = n * m`?

TaylorOshan:
Yah, you're right! I totally tripped myself up on the example I was looking at. [Here is notebook](https://gist.github.com/TaylorOshan/21e3cdfb6ffc855a59fcfb39a8f87b40) to verify. The resulting W is still always square though.

TaylorOshan:
If n=9 and m=16 and assuming there are no self-neighbors and the diagonal is therefore set to 0, then for each of the n*m locations then there would 224 possible neighbor relations  for any (n(i), m(j)) pair. So I think you could create a new n*m W where (n(i), m(j)) = 1 if any of the 224 relations between n(i) and m(j) = 1 in the nm*nm W. I think all you would need is n and m to do so?

sjsrey:
Let's wait on merge until @lanselin gets a chance to look at this.

PR #854, opened 2016-07-27
handle verication context for githubstats

sjsrey:
Something must have changed regarding verification at github since our last release so our script would error out with:

```
 File "/home/serge/anaconda3/envs/pysal2/lib/python2.7/urllib2.py", line 1198, in do_open
    raise URLError(err)
urllib2.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:590)>
```

This is a workaround for that error.


sjsrey:
Rolled into #858

PR #856, opened 2016-07-30
Make the output and build reproducible

sebastic:
As part of the effort to support [Reproducible Builds](https://reproducible-builds.org/) Chris Lamb submitted a patch in [Debian Bug #832997](https://bugs.debian.org/832997) to implement the [`SOURCE_DATE_EPOCH` specification](https://reproducible-builds.org/specs/source-date-epoch/).


sjsrey:
This was tested locally as the failures on travis were due to spreg-mltests hanging.

PR #857, opened 2016-07-30
Fix TabErrors, replace tabs with spaces

sebastic:
Building the GeoPandas Debian package with PySAL failed due to these TabErrors.


sjsrey:
This was tested locally as the failures on travis were due to spreg-mltests hanging.

PR #858, opened 2016-07-31
Dev mltest

sjsrey:
The justification for this PR is to skip ml related tests on travis so the release build can progress. The tests pass locally. An issue will be filed if this pr is merged that reminds us to investigate the ml tests and the reason travis is hanging on them.


ljwolf:
Looks good and Travis likes it, so I think it's okay to merge. 

So weird that the tests are hanging only on Travis...

PR #860, opened 2016-08-16
Incoprate updates to db driver and unittests

ljwolf:
I believe this is the code that disables the warnings in the database driver. 
As I'm closing up the GSOC loose ends, I wanted to make sure this was filed somewhere. 

I believe it's complete:
- When you don't have pandas, no print statement is made on import. 
- When you try to read a file that is registered with the database driver and are missing geomet/shapely/sqlalchemy, an `ImportError` is raised telling you which packages are missing. When a WKB reader is missing:
```
No WKB parser found. Please install one of the following packages to enable this functionality: [geomet, shapely]
```
and when there's no sqlalchemy:
```
No module named sqlalchemy. Please install sqlalchemy to enable this functionality.
```

I've tested it in a clean conda env, with only numpy and scipy.

### DIFF
--- a/pysal/__init__.py
+++ b/pysal/__init__.py
@@ -45,7 +45,6 @@ try:
     from pysal.contrib import pdio
     pysal.common.pandas = pandas
 except ImportError:
-    print('Pandas adapters not loaded')
     pysal.common.pandas = None
     
 # Load the IOHandlers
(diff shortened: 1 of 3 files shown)


sjsrey:
Looks good.

PR #861, opened 2016-08-22
GSOC -SpInt

TaylorOshan:
This PR is to promote review and discussion of what I have produced as a result of my [Google Summer of Code project](https://github.com/pysal/pysal/wiki/SpInt-Development), which was to develop code for spatial interaction modeling and the general analysis of spatial interaction data. 

A list of the current features, and future work is [here](https://github.com/TaylorOshan/pysal/tree/spint/pysal/contrib/spint) and an in-depth log of the project development can be found on the [project blog](https://tayloroshan.github.io/). A jupyter notebook that showcases the use of the main features is [here](https://github.com/TaylorOshan/pysal/blob/spint/pysal/contrib/spint/notebooks/examples/Example_NYCBikes_AllFeatures.ipynb).

Some potential points of discussion:

1. The CountModel class: originally meant to act as a dispatcher to different types of count models and estimation techniques, it currently only dispatched to Poisson models estimated using IWLS, which makes it sort of superficial. If support for more models/estimation techniques are added then this might become more useful. If non-count models (i.e., Gaussian) are added, perhaps a better name reflective of this flexibility should be chosen. 

2. VecMoran class: right now this class relies on list comprehensions to store the output of permutations. For large datasets, the memory overheard becomes huge, so perhaps there is a better way to store only the necessary output rather than say, the entire W object for each permutation. This problem does not arise in a regular Moran's I because the permutation does not effect the spatial organization of the spatial units as it does in the vector-based Moran's I. 

3. OD based weights: the ODW type spatial weight, which is based on the contiguity of both origins and destinations, currently takes in an ``n x n`` origin W and and ``n x n`` destination W (often these are the same) and returns the ``n**2 x n**2`` origin-destination W. It might be useful, depending on how development of a SAR Lag spatial interaction model goes, to have it also return ``n**2 x n**2`` W's for the origins and destinations too. These are just the result of ``np.kron(np.identity(n), originW)`` and ``np.kron(destinationW, np.identity(n))`` and are needed to have W objects that match the dimensions of the ``n x 1`` vector of spatial interaction observations. This point overlaps with #847

4. Default intercept behavior: currently the spatial interaction models are calibrated without intercepts. This can be changed using the boolean parameter ``constant=True``. I found that the intercept needs to be omitted for the results to match those calibrated from maximum entropy and maximum likelihood estimation, though this default behavior might not always be optimal.


sjsrey:
This is an impressive amount of contributions :+1: 

I think the pieces all make sense, even the ones such as the CountModel class where you identify future extensions that maybe will require some thought. That is fine.

One thing that would help is if the https://github.com/TaylorOshan/pysal/tree/spint/pysal/contrib/glm had a README that would let everyone know what this is intended to support. Is it primarily consumed by spint, or could it also be used separately from spint. If the former, maybe it should move into spint?

sjsrey:
We may need to update the [requirements information](https://github.com/pysal/pysal/blob/master/pysal/contrib/README.rst) before merging this.

TaylorOshan:
Yeah, there are no dependencies now, so should I just remove that bit?

sjsrey:
I believe that some of these conflicts were resolved in https://github.com/TaylorOshan/pysal/pull/6#issuecomment-244579172.

TaylorOshan:
Alright, after merging #6 and then making a final edit to the contrib index, I think everything is correct and up-to-date.

Issue #867, opened 2016-09-12
SpInt may cause -inf in input data


jtsmn:
@@##Thank you for filing this issue! To help troubleshoot this issue, please follow
the following directions to the best of your ability before submitting an issue.
Feel free to delete this text once you've filled out the relevant requests. 

Please include the output of the following in your issue submission. If you don't know how to provide the information, commands to get the relevant information from the Python interpreter will follow each bullet point.

Feel free to delete the commands after you've filled out each bullet. 
- Platform information:

```
posix.uname_result(sysname='Darwin', nodename='innsbruck-umh.cs.umn.edu', release='15.5.0', version='Darwin Kernel Version 15.5.0: Tue Apr 19 18:36:36 PDT 2016; root:xnu-3248.50.21~8/RELEASE_X86_64', machine='x86_64')
```
- Python version: 

```
3.4.4 (default, Mar  2 2016, 03:31:27) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)]
```
- SciPy version:

```
0.17.0
```
- NumPy version:

```
1.10.4
```

In the new `spint` implementation on the dev branch, running any model with `cost_func='pow'` will lead to 0s being `-inf`, because the cost_function is set to `np.log` (see `__init()__` in BaseGravity, https://github.com/pysal/pysal/blob/dev/pysal/contrib/spint/gravity.py#L166).

One approach is to use `np.log1p`, which is what is done in the example `.ipynb`

sjsrey:
This seems to be a special case of "self-neighbor". Numerical fixes are possible, but the substantive issues are a bit thornier, so leaving this open.

TaylorOshan:
Hmm, I'm not sure what the best default behavior should be. The error is being thrown because in the Poisson spatial interaction model you still need to log the covariates, and if you have covariates with a value of zero then you will be in trouble. As @jtsmn mentioned one way around it would be to add 1 to values of zero, which is the same as using `np.log1p`, thought I am not aware of any research that looks to see how this ultimately effects the model. It has, however, been shown that adding a small constant to values of zero for the dependent variable can impact the model, so I would assume this is also the case for adding constants to covariates. 

[...]

PR #875, opened 2016-10-25
Spint constant

TaylorOshan:
The justification for this PR is primarily to 

(a) patch some inconsistencies in the constant setting in the SpInt module. Previously, the models operated without a constant by default. This does not affect the main coefficients in the constrained varieties, thought it does change the way the fixed effects used for constraints are interpreted. The impact is greater on the Gravity model, whose parameters change when the constant is added. Now the default setting adds the appropriate constant, thought it could be turned off using ``constant=False``. Tests were updated accordingly to reflect the new behavior.

(b) it also changes the aic attribute to AIC for consistency.

(c) adds value errors when there are values of 0 that would be logged. The errors tell the user specifically which input array the 0 is detected within, including the column. Therefore, this satisfies #867. 

(d) Finally, it adds a pdf (named SpIntPrimer.pdf) with a tutorial on the basic concepts of spatial interaction modeling and how to use them in SpInt. It has been included here because it was written with the changes of (a) in mind and I wanted to make sure it isn't added until the constants settings were consistent. 

Since this is something that has a strong impact on the ``Gravity`` class and given the new rolling release model would this be something that be rolled into master before the accumulation of other new features?


sjsrey:
Definitely something we will pull into the next release. Great work.

PR #876, opened 2016-11-01
changing shp.next() to next(shp)

lincolnfrias:
5. [x] The justification for this PR is: 

`shp.next( )` on python3 gives the error: 
`'PurePyShpWrapper' object has no attribute 'next'.` 
Changing it to `next(shp)` solves the issue.

### DIFF
--- a/doc/source/users/tutorials/fileio.rst
+++ b/doc/source/users/tutorials/fileio.rst
@@ -38,7 +38,7 @@ Shapefiles
 
     >>> import pysal
     >>> shp = pysal.open('../pysal/examples/10740.shp')
-    >>> poly = shp.next()
+    >>> poly = next(shp)
     >>> type(poly)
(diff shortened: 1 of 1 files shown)


ljwolf:
Though I don't think this gets verified in our testing apparatus, it's right to use `next(iterator)` in both 2.X and 3.X since [PEP 3114](https://www.python.org/dev/peps/pep-3114/). 

I've tested locally and both work, so this should probably get merged. 

Stymied about the underlying problem, though. . Not sure how only `FileIO.FileIO.next` gets dropped/clobbered when the `PurePyShpWrapper` appears to correctly inherit all the other method from `FileIO`, like `get` and `flush`. Any insight @schmidtc?

lincolnfrias:
The same thing happens in other places of the documentation, so my PR is not that useful. Thus, I am closing it.

PR #880, opened 2016-11-23
Rel 1.13

sjsrey:
The justification for this PR is: Release 1.13.0


TaylorOshan:
Would it be possible to merge #879 into dev before the release, if it doesn't cause a lot of work? Its not super important as it primarily just moves somes files around so that the SpInt primer is more visible.

PR #902, opened 2017-01-26
Spreg fix

pedrovma:
The justification for this PR is to fix Travis failures on spreg.


sjsrey:
Merging as spreg errors are fixed. Will move onto non-spreg errors.

PR #908, opened 2017-02-07
Geographically weighted regresson

TaylorOshan:
The justification for this PR is to introduce a contributor module for geographically weighted regression (GWR). The GWR module is dependent upon scipy/numpy and the GLM module that is currently in the contrib folder being used by the SpInt module. 

The current GWR code allows for the following options:

 Model types
(1) Gaussian GWR
(2) Poisson GWR
(3) Binomial GWR

Kernel function types
(1) Gaussian (slightly different from than the one in weights)
(2) Exponential
(3) Bi-square

Kernel types
(1) fixed (distance-based)
(2) adaptive (nearest-neighbor)

At the moment, there is a [custom class](https://github.com/TaylorOshan/pysal/blob/gwr/pysal/contrib/gwr/kernels.py#L47) used to compute the inter-observation distances akin to a weights matrix. The reason for this was because the current ``Kernel`` class is optimized for memory overhead, rather than speed and the speed is crucial for GWR. Also, the current class automatically truncates "far" away observations, where this is not always desirable in GWR. A final difference is that this custom class does not return a W object, which was simpler and lighter to work with, since GWR only need the inter-observation distances and not other topological relations. In the future, it could be optimal to rename this and have it somewhere more central in case it is useful for other work.


sjsrey:
Looks good.

PR #913, opened 2017-02-24
0. Fixed docstring that was causing latex error

calebrob6:
The docstring for pysal.esda.mapclassify.Map_Classifier has a small error, see http://pysal.readthedocs.io/en/latest/library/esda/mapclassify.html

This is a small fix. Please let me know if I should open an issue or target dev with this pull request instead.

Thanks!


sjsrey:
Looks good.

Thanks!

PR #919, opened 2017-03-05
Dijkstra mp

jGaboardi:
This PR adds optional multiprocessing functionality to shortest path calculation with `dijsktra_mp()` in `network.utils`. A unit test was also added.

### DIFF
--- a/pysal/network/network.py
+++ b/pysal/network/network.py
@@ -331,7 +331,7 @@ class Network:
 
         return ps.weights.W(neighbors, weights=weights)
 
-    def distancebandweights(self, threshold, processing=None):
+    def distancebandweights(self, threshold, n_proccess=None):
         """
         Create distance based weights
 
@@ -340,15 +340,15 @@ class Network:
         threshold:      float
                         Distance threshold value.
         
-        processing:     int
+        n_processes:    int, str
                         (Optional) Specify the number of cores to utilize.
-                        Default is 1 core.
-                        9999 = use all available cores
+                        Default is 1 core. Use (int) to specify an exact number or cores.
+                        Use ("all") to request all available cores.
         """
         try:
             hasattr(self.alldistances)
         except:
-            self.node_distance_matrix(processing)
+            self.node_distance_matrix(n_proccess)
 
         neighbor_query = np.where(self.distancematrix < threshold)
         neighbors = defaultdict(list)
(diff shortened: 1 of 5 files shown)


ljwolf:
The travis failure looks like it's just on a connectivity issue :upside_down_face: 

I've restarted the CI build. It should pass.

pysal/network/network.py
```diff
@@ -331,19 +331,24 @@ def contiguityweights(self, graph=True, weightings=None):
 
         return ps.weights.W(neighbors, weights=weights)
 
-    def distancebandweights(self, threshold):
+    def distancebandweights(self, threshold, processing=None):
```
ljwolf: Could we call this `n_jobs` or `n_processes` or something?

pysal/network/network.py
```diff
@@ -627,19 +632,50 @@ def enum_links_node(self, v0):
...
+                self.alldistances[node] = (distance, tree)
+                self.distancematrix[node] = distance
+        # Multiprocessing
+        if processing:
+            try:
+                import multiprocessing as mp
+                from itertools import repeat
+            except:
+                raise Execption("Must have multiprocessing and itertools")
+            if processing == 9999:
```
ljwolf: I think it makes more sense to make this a separate type of argument, like `'all'` or `np.inf`, rather than a large int.

pysal/network/network.py
```diff
@@ -627,19 +632,50 @@ def enum_links_node(self, v0):
...
+            for node in self.node_list:
+                distance, pred = util.dijkstra(self, self.edge_lengths, node, n=float('inf'))
+                pred = np.array(pred)
+                #tree = util.generatetree(pred)     <---- something to look at in the future
+                tree = None
+                self.alldistances[node] = (distance, tree)
+                self.distancematrix[node] = distance
+        # Multiprocessing
+        if processing:
+            try:
```
ljwolf: itertools and multiprocessing are part of the [python standard library](https://docs.python.org/2/library/), so this will always evaluate to `True` for CPython implementations.

pysal/network/util.py
```diff
@@ -109,6 +109,65 @@ def dijkstra(ntw, cost, node, n=float('inf')):
...
+    Returns
+    -------
+    distance:   list
+                List of distances from node to all other nodes.
+    
+    pred:       list
+                List of preceeding nodes for traversal route.
+    """
+    
+    v0 = node
```
ljwolf: It looks like this is identical to `dijkstra`, except for the arity of the arguments. Can we get away with:

```python
def dijkstra_mp((ntw, cost, node)):
    return dijkstra(ntw, cost, node)
```
to avoid having the same code in both places?

pysal/network/network.py
```diff
@@ -627,19 +632,50 @@ def enum_links_node(self, v0):
...
+                import multiprocessing as mp
+                from itertools import repeat
+            except:
+                raise Execption("Must have multiprocessing and itertools")
```
ljwolf: oh also this is misspelled

Issue #812, opened 2016-05-26
Add a WKT/WKB __repr__/__str__ methods to pysal polygons & chains


ljwolf:
Our point geometries already implement something close to the WKT standard:

``` pysal
>>> pysal.cg.Point((10,5))
(10.0, 5.0)
```

writing this shouldn't be too onerous. But, if in place, it would allow us to do something like:

```
>>> df = ps.pdio.read_files('NAT.shp')
>>> df['new_column'] = do_spatial_analysis()
>>> df.to_csv('NAT.csv')
>>> data = pandas.read_csv('NAT.csv')
>>> data['geometry'] = data.geometry.apply(pysal.core.utils.WKTParser.Polygon)
```

which could be quite nice. Currently, this is the type of behavior [adopted in shapely](https://github.com/Toblerity/Shapely/blob/master/shapely/geometry/base.py#L227), where shapes's `__str__` method provides wkt and their `__reduce__` method (used by pickle) provides wkb.

parulsethi:
Hi,
I am interested in adding this, but I'm not much familiar with pysal workflow. So, could you please elaborate this a bit more, maybe an example about what output is required for polygon/chain. Like, what resultant format of string representation for polygon?

ljwolf:
Hey! Thanks for your interest!

This involves providing a correct mapping from our polygon type (in `pysal.cg.shapes`) to the Open Geographic Consortium's [Well-known text standard for representing geometries](https://en.wikipedia.org/wiki/Well-known_text#Geometric_objects). 

IIRC, the blocker for this was that our polygon type, `pysal.cg.shapes.Polygon`, did not preserve the mapping for which holes nest within each exterior ring. All the other shape types with corresponding types in WKT can be implemented using a string template or regular expressions.  

In order to generate valid WKT for multipolygons with holes, each exterior polygon ring in `polygon.parts` needs to get paired with the holes in `polygon.holes` contained inside of that exterior ring. This could be done by checking whether a point on the boundary of each hole is within each exterior ring using `Polygon.contains_point`, assuming the polygon boundary does not intersect itself.

parulsethi:
Thanks for the comprehensive explanation!

and if I understand correctly, I'd need to check only a single point out of a hole's vertices, rather than checking for all vertices of a hole to be inside the exterior ring, as no intersection is being assumed

Issue #852, opened 2016-07-24
`contains_point` cannot handle nested exterior rings.


ljwolf:
I hit this bug working on the WKT serialization/deserialization for the labelled array gsoc.

Take the following example with two polygons. The first multipolygon is a double-torus defined by a square ring between 1 and .75, and another square ring between .5 and .25. The other multipolygon is a single torus with an island, defined by a square between 1 and .75, and an inner square at .5. 

Our `contains_point` implementation bails when any hole contains the point, which is accurate for the double-torus, but incorrect for the single-torus-with-island:

```python
>>> import pysal as ps
>>> double_torus = ps.cg.Polygon(parts= [[(-1, -1), (-1,1),(1,1),(1,-1)],  #exterior square at 1
                                        [(-.5,-.5), (-,5, .5),(.5,.5),(.5,-.5)]],  #exterior square at .5
                                 holes= [[(-.75, -.75),(-.75,.75),(.75,.75),(.75,-.75)], #hole square at .75
                                        [(-.25,-.25),(-.25,.25),(.,25, .25),(.25,-.25)]]) #hole square at .25
>>> single_island_in_torus =  ps.cg.Polygon(parts= [[(-1, -1), (-1,1),(1,1),(1,-1)],  #exterior square at 1
                                        [(-.5,-.5), (-,5, .5),(.5,.5),(.5,-.5)]],  #exterior square at .5
                                 holes= [[(-.75, -.75),(-.75,.75),(.75,.75),(.75,-.75)], #hole square at .75
                                         ]) #inner square is solid
>>> double_torus.contains_point((0,0)) #is hollow, so origin is not contained
False
>>> single_island_in_torus.contains_point((0,0)) #should contain origin, since the inner square is solid
False
```

Again, if [*any* hole contains the point, the algorithm bails](https://github.com/pysal/pysal/blob/master/pysal/cg/shapes.py#L1329), meaning any concentric multi-polygon will have incorrect results for point-in-polygon searches. 

In fact, I don't think we can do a correct point in polygon search on multipolygons without establishing a ring-hole nesting. 

Like, take a point P related to a multipolygon M composed of 2 exterior rings, A,B and one hole, H. Let P be contained in B. Stating the rings in OGC style, if M := ((AH), B), then P in H is not sufficient to exclude P from M, since P in (B intersection H) implies P in M. 

[...]

PR #920, opened 2017-03-06
added wkt representation for chain/polygon

parulsethi:
Hi,
This PR is to address the [issue-812](https://github.com/pysal/pysal/issues/812).

 It adds `__repr__()` method for Chain and Polygon following convention similar to [this](https://en.wikipedia.org/wiki/Well-known_text#Geometric_objects). Please suggest for anymore changes/additions required.

### DIFF
--- a/pysal/cg/shapes.py
+++ b/pysal/cg/shapes.py
@@ -11,6 +11,7 @@ import math
 from warnings import warn
 from sphere import arcdist
 import numpy as np
+import copy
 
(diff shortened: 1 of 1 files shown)


ljwolf:
Hi! This looks good for simple polygons. 

An [issue](https://github.com/pysal/pysal/issues/852) that needs consideration for this solution is when a polygon has a hole in a polygon in a hole, figure (d) in the linked issue. 

In that case, a simple point-in-polygon check will not correctly assign the hole to the right containing exterior ring.

Issue #813, opened 2016-05-26
w.remap_ids(ids) never sets w.id_order_set


ljwolf:
I found this while working on the polymorphic weights constructors: 
```python
import pysal as ps
NATpath = ps.examples.get_path('NAT.shp')
w = ps.queen_from_shapefile(NATpath)
print(w.id_order_set) # correctly prints False
w2 = ps.queen_from_shapefile(NATpath, idVariable='FIPS')
w2.id_order_set # should be True, but prints False
```
This isn't really a bug with the `*_from_shapefile` functions, because the `*_from_shapefile` constructors typically look like this:
```python
def weights_from_shapefile(collection, sparse=False, ids=None):
    w = do_weights(collection)
    if sparse:
        w = make_sparse(w, ids=ids)
    if ids is not None:
        w.remap_ids(ids)
    return w
```
so, `remap_ids` needs to change the flag when it finishes in order for these to consider the id order to be set. I think this might also have some interaction with #799, but I'm unsure.

The fix for this should be straightforward.

ljwolf:
or maybe I'm misunderstanding what `id_order_set` means. Does it only get set if the desired iteration order is not the order in which the ids are passed?

sjsrey:
I see the source of confusion. `id_order_set` was intended to signify that the ordering was changed to be different from that reflected in the order of the shapes in the shapefile.

`remap_ids` is used to keep the same order as the shapes in the shapefile, but use a different attribute to label the shapes (other than ints going from 0 to n-1).

So the answer to the question by @ljwolf above is "Yes, the intention was to use `id_order_set` only when the desired iteration order is changed. 

Does this clarify?

ljwolf:
Hmm, I'll think about this and raise any further questions here. If this is intended, then that's fine. 

I'm still a little confused by the combination of `id_order`, `ids`, and how separately using `remap_ids(new_ids)` vs. `w.id_order = new_ids` works. 

The cause of this concern was that, if I wanted a `W` in the same order as a dataframe, whether to use `ids` or `id_order` as the keyword argument. So, going forward, I'll use `id_order` to mean "use this column of the dataframe and treat it like it's ordered`, and `ids` as "use this column of the dataframe as if it were not ordered.

PR #922, opened 2017-03-14
Weights, ordering, and from_dataframe

ljwolf:
The geotable API for Contiguity weights [was somewhat confusing](https://github.com/pysal/pysal/issues/813) to keep consistent with the preexisting file API. In doing so, I chose to use an indexing scheme matching the pre-existing `weights.util.queen_from_shapefile`, in that neither `ids` nor `idVariable` caused the resulting `W` to have any special ordering. 

Later, working with this stuff, I got surprised by some instability in the ordering of `W` when the indexing column is a string. Since we force `id_order` to be sorted, if `df` is **not** sorted, then `from_dataframe(df, idVariable='my_string_index')` will not be aligned. 

So, this PR does a few things. First, it breaks out the functionality for classmethods in `Contiguity` to live in a [mix-in](https://en.wikipedia.org/wiki/Mixin) class, `_ContiguityMixin`.  Then, the `from_dataframe` method is changed to ensure that the alignment of `W` matches `df` unless you explicitly ask for it not to be aligned with `id_order = False`. Unittests cover:
- `from_dataframe(df, idVariable='my_index')` is ordered by `my_index` when `my_index` is a `str`, `int`, or `float`
- `from_dataframe(df, ids=df.my_index.tolist())` is also ordered over `my_index` when `my_index` is a `str`, `int`, or `float` 
- an explicit `id_order=False` flag must be passed to construct a `W` that is out of alignment with the source data frame.
- Alignment can be assured after `df.sort_values('my_index')` for `str`, `int`, `float`.

### DIFF
--- a/pysal/weights/util.py
+++ b/pysal/weights/util.py
@@ -521,14 +521,17 @@ def higher_order_sp(w, k=2, shortest_path=True, diagonal=False):
 
     """
 
-    tw = type(w)
     id_order = None
-    if tw == pysal.weights.weights.W:
-        id_order = w.id_order
-        w = w.sparse
-    elif tw != scipy.sparse.csr.csr_matrix:
-        print "Unsupported sparse argument."
-        return None
+    if issubclass(type(w), pysal.weights.W):
+        if np.unique(np.hstack(w.weights.values())) == np.array([1.0]):
+            id_order = w.id_order
+            w = w.sparse
+    elif scipy.sparse.isspmatrix_csr(w):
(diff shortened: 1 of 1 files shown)


ljwolf:
I definitely think I clobbered this accidentally. Something conflicted my my laptop, so I picked the recent changes in attempting to address #944 and this got overwritten. I'm not sure how to salvage, and at this point, I don't even recall what the changes were in this PR... 

shoot.

PR #925, opened 2017-03-28
Reorg

sjsrey:
Start of Makefile for exploring reorg_dist.

Note that for the modules we are bringing in (only libpysal and esda for exploration at the moment) we have to haul their package folders up one level after cloning. This works but I'm wondering if there are going to be problems induced here? 

We could punt on the source redistribution and just rely on pip to install the dependencies (i.e., libpysal, esda, spint.......)? In this case the makefile could be used for setting up a local dev environment, while the dependences remain outside of the pysal/pysal upstream repository? Food for thought.

Also added a setup.py for the metapackage.  This would have pip do the heaving lifting on dependencies - pysal becomes a lightweight metapackage.


ljwolf:
test failures look like an issue in the `libpysal` API.

PR #926, opened 2017-03-30
DOC: fix a broken link (HTTP 404) in documentation

yogabonito:
A link in PySAL's documentation was broken (404 Not Found). With this PR it will point to an existing document again.


yogabonito:
Thanks for your suggestion @sjsrey! I changed the link accordingly.

PR #947, opened 2017-07-05
[WIP] fix bug #944

ljwolf:
This adds "correct" typedispatching for the `higher_order` function.

It should work for any weights subclass of `W` that's binary, or on binary sparse matrices.

It requires unittests to ensure all the W-types complete for it.


ljwolf:
this fails because of some stuff in locators that breaks at a numpy bump. I thought I fixed this? Maybe it's in libpysal.

PR #957, opened 2017-07-12
Fix minor bugs in tests

karldw:
- test_sphere.py:
Existing code was trying to compare lists of floats with `assertAlmostEqual` instead of the numpy equivalent. 

- test_examples.py:
Existing code assumed a "/" path separator, so failed on Windows.
The change from `pathparts[-3:]` to `pathparts[-2:]` is because `os.path.normpath` strips out the trailing slash.

### DIFF
--- a/pysal/cg/tests/test_sphere.py
+++ b/pysal/cg/tests/test_sphere.py
@@ -3,7 +3,7 @@ import pysal
 import math
 import unittest
 import numpy as np
-
+from pysal.common import (RTOL, ATOL)
 
 class Sphere(unittest.TestCase):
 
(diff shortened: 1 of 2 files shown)


ljwolf:
Hey! thanks for this PR. The stuff in test examples is :+1: 

For `self.assertTrue(np.allclose(...))`, we have been rewriting tests to use `np.testing.assert_allclose` and the package-wide constants `RTOL` and `ATOL` in `pysal.common`.

karldw:
The [unittest docs recommend](https://docs.python.org/2.7/library/unittest.html#basic-example) using `assertTrue()` over the built-in `assert` "so the test runner can accumulate all test results and produce a report." I didn't realize nose was more sophisticated.

PR #966, opened 2017-07-27
Preserving more than one underscore if present in the original column name. Currently if a column name has more than one underscore, they are smashed into one.

nikisix:


ljwolf:
Hey! Thanks for this! will incorporate into [pysal/esda](https://github.com/pysal/esda)

Issue #970, opened 2017-08-29
Odd pysal.Moran() randomisation VI


rsbivand:
There appear to be issues in pysal.Moran() in VI_rand and derived values. The final - (EI)^2 appears to be missing (present in VI_norm), but in addition the kurtosis term is tiny. The VI is hard to reconstruct in R, but VI_rand in pysal.Moran() is completely different from all others I've seen (CrimeStat, multiple R implementations). Grateful for insights.

sjsrey:
@rsbivand thank you for reporting this. It is indeed a bug as the formula for V(I)_rand was implemented incorrectly. We are correcting and adding test coverage.

rsbivand:
Thanks! Reprex:

```
library(reticulate)
pysal <- import("pysal")
np <- import("numpy")
pth <- pysal$examples$get_path("stl_hom.txt")
f <- pysal$open(pth)
col <- f$by_col("HR8893")
y <- np$array(col)
w <- pysal$open(pysal$examples$get_path("stl.gal"))$read()
mi <- pysal$Moran(y, w, two_tailed=FALSE)
df <- read.csv(pth, skip=1)
library(spdep)
nb <- read.gal(pysal$examples$get_path("stl.gal"))
miR <- moran.test(df$HR8893, nb2listw(nb, style="W"))
all.equal(mi$I, unname(miR$estimate[1]))
all.equal(mi$EI, unname(miR$estimate[2]))
all.equal(mi$VI_rand, unname(miR$estimate[3]), scale=1)
all.equal(mi$z_rand, unname(miR$statistic), scale=1)
miN <- moran.test(df$HR8893, nb2listw(nb, style="W"), randomisation=FALSE)
all.equal(mi$VI_norm, unname(miN$estimate[3]), scale=1)
all.equal(mi$z_norm, unname(miN$statistic), scale=1)
```

rsbivand:
Please let me know when it reaches master, or say how to access the patch - I can see it, but can't yet pull.

rsbivand:
Thanks! Looks good - the write-up of the comparison will give the > 1.14.1 output for Moran & Geary, and note in text that <= 1.14.1 are off for randonisation.

rsbivand:
Confirmed both global Moran VI_rand and Geary VC_rand now equal within machine precision of R/spdep implementation, thanks!

PR #972, opened 2017-08-31
Bug Fix for #970

sjsrey:
Correcting variance of Moran's I under randomization.

Adding test cases.


pysal/esda/moran.py
```diff
@@ -129,6 +129,16 @@ class Moran(object):
     >>> mi_1.p_norm
     5.7916539074498452e-05
 
+    Example from http://www.lpc.uottawa.ca/publications/moransi/moran.htm
```
ljwolf: This is good that we test it here, but iirc, we don't have doctests running on travis? 

Is this something we want to turn on?

ljwolf: I guess since it's got both a unittest & a doctest, only running the unittest is fine. Disregard.

pysal/esda/moran.py
```diff
@@ -188,13 +198,18 @@ def __moments(self):
...
-        self.seI_rand = vi ** (1 / 2.)
+        # variance under randomization
+        xd4 = z**4
+        xd2 = z**2
+        k_num = xd4.sum() / n
+        k_den = (xd2.sum() / n)**2
+        k = k_num / k_den
+        EI = self.EI
+        A = n * ((n * n - 3 * n + 3) * s1 - n * s2 + 3 * s02)
+        B = k * ((n * n - n) * s1 - 2 * n * s2 + 6 * s02  )
```
ljwolf: these are pretty hairy terms. The only things I think would be more readable here are to always refer to `n * n` as `n**2` and use `n*(n-1)` for `(n**2 - n)`, but that latter one may just be my own perception. 

Regardless, this looks okay as merged. I'm not sure the original reference on the variance under randomization though? Would that be in the Tiefelsdorf exact distribution paper?

pysal/esda/tests/test_moran.py
```diff
@@ -25,6 +25,13 @@ def test_sids(self):
         np.testing.assert_allclose(mi.I, 0.24772519320480135, atol=ATOL, rtol=RTOL)
         self.assertAlmostEquals(mi.p_norm,  5.7916539074498452e-05)
 
+    def test_variance(self):
+        y = np.arange(1, 10)
+        w = pysal.lat2W(3, 3)
+        mi = pysal.Moran(y, w, transformation='B')
+        self.assertAlmostEquals(mi.VI_rand, 0.059687500000000004)
+        self.assertAlmostEquals(mi.VI_norm, 0.053125000000000006)
```
ljwolf: should this use allclose + rtol/atol from `pysal.common`?

pysal/esda/moran.py
sjsrey: Good catches. Will update.

The original reference on the moments is Cliff and Ord (1981) . The link in our doctest has a worked example that reproduces those moments.

sjsrey:
Testing the fix locally against the [reprex](https://github.com/pysal/pysal/issues/970) and all seems to replicate spdep.

PR #973, opened 2017-09-01
Passing p_thres in plot_lisa_cluster. This fixes #971

### LINKED ISSUE #971: base_lisa_cluster function call does not pass p_thresh
The calling function plot_lisa_cluster(shp_link, lisa, p_thres=0.01...) in mapping.py does not not pass the p_thres argument to 
 base_lisa_cluster(lisa_obj, lisa) function call on line 1100

I think it should be:
base_lisa_cluster(lisa_obj, lisa, p_thres)

https://github.com/pysal/pysal/blob/1e254a023ca40949ce7c6f52421726495958f71e/pysal/contrib/viz/mapping.py#L1100

darribas:
Minor patch to make sure `p_thres` is passed on `plot_lisa_cluster`. Note this API is on its way out as it does not depend on `geotable` structures, but it is anyway good to have it properly working.

### DIFF
--- a/pysal/contrib/viz/mapping.py
+++ b/pysal/contrib/viz/mapping.py
@@ -1097,7 +1097,7 @@ def plot_lisa_cluster(shp_link, lisa, p_thres=0.01, shp_type='poly',
     shp = ps.open(shp_link)
     # Lisa layer
     lisa_obj = map_poly_shp(shp)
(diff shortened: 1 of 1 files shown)


sjsrey:
Merging. Test failure is due to travis timing out.

PR #983, opened 2017-09-25
Fix inconsistent use of tabs and spaces in indentation.

sebastic:
Causes TabError:
```
Sorry: TabError: inconsistent use of tabs and spaces in indentation (iwls.py, line 56)
Sorry: TabError: inconsistent use of tabs and spaces in indentation (iwls.py, line 56)
```

The issue was reported by Andreas Beckmann in [Debian Bug #876700](https://bugs.debian.org/876700)

PR #984, opened 2017-09-25
Level dev & master, incorporate 983

ljwolf:
Last release, we just merged directly into master for simplicity. 

If we want to just keep doing that, this should be closed, docs updated, and #983 merged. 

Otherwise, merge this to update dev with all of what went into master for 1.14.2, and add @sebastic #983 to the development head.


ljwolf:
I should say, I'm game to merge directly into master since we have ongoing CI.

sjsrey:
I'm +1 on doing prs into master and dropping dev. Causes more problems than it solves now.

PR #993, opened 2017-10-29
B991

ljwolf:
This resolves bugs 991 & 992. In addition, it merges in some work that entered pysal's development branch by mistake. 

Further, it clarifies & reinforces in the PR template that we are submitting against the master branch.


MalithDis:
I got error this error 
 ImportError: cannot import name 'chisqprob'
when run below code. how to solve this. please help me.

```
#Import the necessary Python moduless
import pandas as pd
import geopandas as gpd
import numpy as np
from geopandas.tools import sjoin
import folium
from folium.plugins import MarkerCluster
from folium.element import IFrame
import shapely
from shapely.geometry import Point
import unicodedata
import pysal as ps
```

PR #996, opened 2017-11-02
edit README for example datasets

weikang9009:


weikang9009:
I did not add a README.md to several new datasets, including *clearwater*, *georgia*, *londonhp*, *nyc_bikes*, *Polygon_Holes*, and *tokyo*. Since people who added them know the datasets better, it would be best if they could add a README.md

PR #989, opened 2017-10-12
resolve discrepancy of estimating Moran's I corrected for rates between GeoDa and pysal

weikang9009:
This PR is to resolve the discrepancy in estimating EBI (Moran's I corrected for rates) between GeoDa and pysal as pointed out in #987. Now users could choose to conform with Geoda by setting parameter `geoda_rate=True` :
```python
pysal.esda.moran.Moran_Local(e, b, w, adjusted=True, transformation = "r", geoda_rate=True)
pysal.esda.moran.Moran_Local_Rate(e, b, w, adjusted=True, transformation = "r", geoda_rate=True)
```
or to use the original implementation of correcting rates in pysal which conform with the Assunção's paper by setting parameter `geoda_rate=False":

```python
pysal.esda.moran.Moran_Local(e, b, w, adjusted=True, transformation = "r", geoda_rate=False)
pysal.esda.moran.Moran_Local_Rate(e, b, w, adjusted=True, transformation = "r", geoda_rate=False)
```

Issue #987, opened 2017-10-02
discrepancy in EB Local Moran's I with PySAL vs. GeoDa results


timstallmann:
Hi! I'm reviewing some results from a student RA and trying to cross-check PySAL output against GeoDa output. I'm finding that for the Local Moran's I with EB Rate calculations, the PySAL output doesn't match up with GeoDa and in some cases there is even a discrepancy in the sign of the calculated Moran's I value. Standard Moran's I results are identical.

Here's the code I'm using (which pulls a base path from the first argument to the script, runs standard and EB-corrected Local Moran's I, and outputs the results to a CSV). 
[Code](https://gist.github.com/timstallmann/07b88cd3cc524e83c6699b77294fbb54)
[Shapefile](https://github.com/pysal/pysal/files/1350223/test.zip)

Results I'm seeing:

geoid | morans_i_eb | morans_i_standard | geoda_morans_i_eb | geoda_morans_i_std
-- | -- | -- | -- | --
370370201031.00 | 0.289457818 | 0.183617922 | 0.2568157 | 0.1836179
370370201032.00 | 0.007694115 | 0.041327785 | 0.0332904 | 0.0413278
370370201041.00 | -0.062648125 | -0.092541534 | -0.0211115 | -0.0925415
370370201042.00 | 0.248481047 | 0.585390201 | 0.359368 | 0.5853902
370370201051.00 | 0.547539092 | 0.392660392 | 0.5357633 | 0.3926604
370370201052.00 | 0.20509518 | 0.18996257 | 0.2376427 | 0.1899626
370370201061.00 | 0.775430212 | 1.205775439 | 0.8670587 | 1.2057754
370370201062.00 | 0.363722315 | 0.267114708 | 0.3676368 | 0.2671147
370370201063.00 | 0.558066514 | 0.64418037 | 0.5816549 | 0.6441804
370370202001.00 | 1.180070252 | 1.221176998 | 1.2955742 | 1.221177
370370202002.00 | 0.201854063 | 0.343386909 | 0.2345107 | 0.3433869
370370202003.00 | -0.219698489 | 0.024902698 | 0.0479546 | 0.0249027
370370202004.00 | -0.026982976 | -0.033460539 | -0.062049 | -0.0334605
370370203001.00 | -0.053438527 | 0.013403188 | -0.094403 | 0.0134032
370370203002.00 | -0.123336382 | -0.068843153 | -0.1495362 | -0.0688432
[...]

PR #1011, opened 2017-11-21
resolve discrepancy of estimating Moran's I corrected for rates between GeoDa and pysal

weikang9009:
This PR is to reopen PR #989 which was unmerged but closed because of some complication I made. It is to resolve the discrepancy in estimating EBI (Moran's I corrected for rates) between GeoDa and pysal as pointed out in issue #987. Now users could choose to conform with Geoda by setting parameter `geoda_rate=True` :
```python
pysal.esda.moran.Moran_Local(e, b, w, adjusted=True, transformation = "r", geoda_rate=True)
pysal.esda.moran.Moran_Local_Rate(e, b, w, adjusted=True, transformation = "r", geoda_rate=True)
```
or to use the original implementation of correcting rates in pysal which conform with the Assunção's paper by setting parameter `geoda_rate=False`:

```python
pysal.esda.moran.Moran_Local(e, b, w, adjusted=True, transformation = "r", geoda_rate=False)
pysal.esda.moran.Moran_Local_Rate(e, b, w, adjusted=True, transformation = "r", geoda_rate=False)
```

### DIFF
--- a/doc/source/developers/release.rst
+++ b/doc/source/developers/release.rst
@@ -37,13 +37,13 @@ As of version 1.6, docs are automatically compiled and hosted_.
 Make a source dist and test locally (Python 3)
 ----------------------------------------------
 
-On each build machine::
+On each build machine
 
   $  git clone http://github.com/pysal/pysal.git
   $  cd pysal
   $  python setup.py sdist
-  $  cp dist/PySAL-1.14.2.tar.gz ../test/.
-  $  cd ../test
+  $  cp dist/PySAL-1.14.2.tar.gz ../junk/.
+  $  cd ../junk
   $  conda create -n pysaltest3 python=3 pip
   $  source activate pysaltest3
   $  pip install PySAL-1.14.2.tar.gz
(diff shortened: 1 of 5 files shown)


sjsrey:
Adding a note here that when this gets merged, we need to port to https://github.com/pysal/esda

weikang9009:
@sjsrey If adjusted=True (or `geoda_rate=False` in function ```assuncao_rate``` and related Moran and local Moran classes) , conform with Assuncao and Reis (1999), that is, assign v_i = a+b/x_i and check individual v_i: if v_i<0, assign v_i = b/x_i.

sjsrey:
@weikang9009 if adjusted=False, is geoda_rate ignored?

weikang9009:
@sjsrey yes. If ```adjusted=False```, ```geoda_rate``` is ignored. Moran statistics won't be adjusted for rates.

sjsrey:
@weikang9009 maybe we should add to the docs that:

```If adjusted=False, geoda_rate is ignored```

weikang9009:
@sjsrey Thank you for the suggestion. Have updated the docs (for `Moran_Rate` and `Moran_Local_Rate`).

PR #1029, opened 2018-03-31
show both figures in mapping.py

slumnitz:
fixed return from plot_lisa_cluster, returns figure now instead of NONE, removed plt.show()
fixed return from plot_cloropleth, returns figure now instead of NONE, removed plt.show()

added plt.show() in the end and redefined LISA and Cloropleth as fig1 and fig2, enabled display of both figures simultaneously.

I also started to make some pep8 fixes


darribas:
This looks good to me!

PR #1031, opened 2018-04-10
Correct bbcommon() docstring

wtadler:
The docstring for bbcommon() indicates that an output of one indicates that the bounding boxes do not overlap. However, it actually indicates that they *do* overlap. This corrects the docstring.


sjsrey:
Good catch.

PR #1047, opened 2018-07-16
ENH: Add spaghetti to meta-package

sjsrey:


jGaboardi:
@sjsrey will #1047 work without spaghetti being up on pypi?

sjsrey:
@jGaboardi yes it will as this is a source distribution that is pulling from master on github not pypi.

.travis.yml
```diff
@@ -34,6 +34,9 @@ script:
   - python setup.py sdist >/dev/null
   - echo "check_stable=False" >pysal/config.py
   - travis_wait 45 nosetests --verbose --with-coverage --cover-package=pysal;
+  - nosestests --verbose --with-coverage pysal/dynamics
+  - nosestests --verbose --with-coverage pysal/explore
```
ljwolf: I would like to check if `pytest` does this right.

sjsrey: Worth a shot. If pytest isn't the solution, we can always go with the explicit approach here.

PR #1063, opened 2018-11-17
Update of convert script and resulting source

sjsrey:
Moving this upstream so others may inspect and help polish.

We need to do some further testing of 

- spaghetti
- pointpats

with regard to how they work with the new import scheme in the meta package


sjsrey:
Note sure why this fails to import fileio?

Locally if I build with `python setup.py sdist` and then create a clean conda environment and then pip install that tarball, the imports work.

weikang9009:
image
the name of the module is not changed. Should be lower case.
image

sjsrey:
@weikang9009 Thanks for spotting this! 

This is what I get for programming on a Mac (case insensitivity).

jGaboardi:
@sjsrey is there anything specific I can help with in regards to `spaghetti`?

sjsrey:
@jGaboardi I'm working on automatically rewriting the imports for spaghetti when it is included as a package within the metapackage. You can see the changes [here](https://github.com/pysal/pysal/pull/1063/commits/3f54a7bc4745851cdd20568cf93c99db805e5ba2)

ljwolf:
I see this fails on spvcm stuff pysal -> libpysal stuff. Thought this was done... sorry. Will fix.

release.md
```diff
@@ -0,0 +1,612 @@
...
+* esda needs an api.py module [(#9)](https://github.com/pysal/esda/issues/9)
+* chore: Update setup for 3+ [(#15)](https://github.com/pysal/esda/pull/15)
+* Master [(#10)](https://github.com/pysal/esda/pull/10)
+
+### giddy:
+* remove giddy.api in README.rst [(#66)](https://github.com/pysal/giddy/pull/66)
+*  chore: update for libpysal lower case module name changes [(#65)](https://github.com/pysal/giddy/pull/65)
+* remove api.py [(#62)](https://github.com/pysal/giddy/issues/62)
+* set up travis dual testing against mapclassify and  esda [(#63)](https://github.com/pysal/giddy/issues/63)
+* replace `libpysal.api` imports with new imports in `markov.py` and `d… [(#61)](https://github.com/pysal/giddy/pull/61)
```
jGaboardi: seems to have an extra "`" here which is causing formatting to get out of whack.

sjsrey: Yup. But I think still  [renders ok](https://github.com/sjsrey/pysal/blob/next/release.md)

PR #1071, opened 2019-01-14
Rel2.0

sjsrey:
This would move master over to 2.0.

We would then have three branches:

- master: stable six-month release
- next: updates of meta-package for next 6-month release
- legacy: 1.14.4 which would be used for bug fixes only


build.md
```diff
@@ -1,13 +1,13 @@
 # Building pysal source release
 
-As of version 2.0 PySAL has become a meta-package that combines a family of spatial analysis packages into a single source distribution.
+As of version 2.0.0 PySAL has become a meta-package that combines a family of spatial analysis packages into a single source distribution.
 
 This document provides instructions on the preparation of the meta-package
```
jGaboardi: Add punctuation.

> This document provides instructions on the preparation of the meta-package.

PR #1078, opened 2019-01-30
DOC: readthedocs files

sjsrey:
These were missing so RTD is not building correctly.

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,0 +1,7 @@
+python:
+    version: 3.5
+    setup_py_install: true
+    pip_install: true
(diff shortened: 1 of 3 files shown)


readthedocs.yml
```diff
@@ -0,0 +1,7 @@
+python:
+    version: 3.5
+    setup_py_install: true
+    pip_install: true
+    extra_requirements:
+        - tests
+        - docs
```
weikang9009: docs requirements are not read. Should add sth in setup.py https://github.com/pysal/pysal/blob/master/setup.py#L48

PR #1080, opened 2019-02-04
Remove executable bit from test & example data files.

sebastic:
The lintian QA tool reported that several data files used for tests have the executable bit set.

This PR fixes the permissions for test & example data.


sjsrey:
We have to move this upstream to libpysal.

sjsrey:
Thanks for this. Will include in the 6-month release.

PR #1093, opened 2019-04-27
doc: Create dash (zeal) docs

sjsrey:
This will create PySAL docsets for [dash](https://kapeli.com/dash) and  [zeal](https://github.com/zealdocs/zeal).

Zeal example:

Screenshot from 2019-04-27 15-21-59


sjsrey:
> These new docs look nice. Are we moving away from the Jekyll model from [pysal.github.io](https://github.com/pysal/pysal.github.io)?

These are a complement to the on-line docs - zeal and dash are offline apps that can be used to access dockets.

Issue #69, opened 2013-02-19
Replace W._characteristics with separate properties


sjsrey:
_Original author: schmi...@gmail.com (April 23, 2010 23:15:06)_

Separate the_characteristics into separate properties to increase flexibility and speed.

_Original issue: http://code.google.com/p/pysal/issues/detail?id=69_

PR #1115, opened 2019-07-16
ENH: adding notebooks to source release

sjsrey:
We probably need to add some testing of the notebooks with the rewrites. This is at least a start of the structure intended to work with the [notebooks project](https://github.com/pysal/notebooks).


darribas:
I'm going to merge for now so I can take a look. Something I want to look into is to add infrastructure to easily run (locally) all notebooks within the `notebooks` project (tracked on [`#69`](https://github.com/pysal/notebooks/issues/69)). This would allow us to catch easy bugs locally before we release a new version of the notebooks website.

PR #1121, opened 2019-07-24
BUG: add zstd as a dependency to work around conda 

sjsrey:
https://github.com/pysal/libpysal/pull/168


sjsrey:
> Should something like this be done for `spaghetti`, as well?

If fiona is a dep, then the current bug would have been raised for spaghetti.

Issue #1125, opened 2019-07-29
BUG: Travis error


sjsrey:
Something seems to have changed in the stack as travis is now [failing to build](https://travis-ci.org/pysal/pysal/jobs/565093013#L827). 

Following the [guidance at travis](https://docs.travis-ci.com/user/common-build-problems/#my-tests-broke-but-were-working-yesterday), I went back to a build that had passed, and restarted it. Sure, enough that one now errors for the same reason.

jGaboardi:
So this is a "wait and see what happens" situation for the moment?

sjsrey:
I think the culprit might be [libgfortran-ng-7.3.0](https://travis-ci.org/pysal/pysal/jobs/565093013#L687) as it is not fully downloading.

jGaboardi:
While looking into this I found that it appears [we are not actually testing with Python 3.7](https://github.com/pysal/pysal/issues/1127).

jGaboardi:
(for searchable error documentation purposes)

> [Yeah, seems like it because each attempted build fails with around 75% of libgfortran downloaded. However, I haven't found any other projects/builds that are currently failing.](https://github.com/pysal/pysal/issues/1127#issuecomment-516448332)


```
libgfortran-ng-7.3.0 | 1006 KB   | ###########################8          |  75%
```

```
ReadError('file could not be opened successfully')
The command "conda env create --file="${ENV_FILE}"" failed and exited with 1 during .
Your build has been stopped.
```

sjsrey:
[conda-forge hint](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=57635&view=logs&jobId=1bf226d3-0e2f-52d8-fa93-7d9e633347b3&taskId=ec8d466d-e3b4-5115-4f06-222398f91e6c&lineStart=3139&lineEnd=3140&colStart=1&colEnd=1)

jGaboardi:
```
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
export PREFIX=/home/conda/feedstock_root/build_artifacts/pysal_1564424786612/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_
export SRC_DIR=/home/conda/feedstock_root/build_artifacts/pysal_1564424786612/test_tmp
import: 'pysal'
[...]

PR #1128, opened 2019-07-30
Fix Travis-CI

ocefpaf:
This is passing on my Travis-CI branch, see https://travis-ci.com/ocefpaf/pysal/builds/121305427.

I guess that just the `conda update` was enough but the rest is nice to have. In particular the `strict` channel configuration when using conda-forge.


jGaboardi:
Resolves #1125

PR #1130, opened 2019-07-31
Don't use conda-env files with Travis-CI

ocefpaf:
I'll explain the details inline but the main change here is to avoid using conda-env environment files but use the same requirements files to instead to create the test env. This will prevent mismatch between the testing env and what people will get when installing `pysal` with wither `pip` or `conda`.


.travis.yml
```diff
@@ -1,13 +1,18 @@
-language: python
+language: minimal
```
ocefpaf: If using `conda` for testing we don't need the image with `Python`.

.travis.yml
```diff
@@ -1,13 +1,18 @@
...
+  fast_finish: true
   include:
-    - env: ENV_FILE="ci/travis/36-latest-conda-forge.yaml"
-    - env: ENV_FILE="ci/travis/37-latest-conda-forge.yaml"
+    - name: "python-3.6"
+      env: PY=3.6
+    - name: "python-3.7"
+      env: PY=3.7
+    - name: "tarball"
+      env: PY=3.7
```
ocefpaf: This test matrix implements the two tests that existed, for 3.6 and 3.7, and add a new one to check the `tarball` created. This will help prevent some release issues.

.travis.yml
```diff
@@ -18,17 +23,22 @@ install:
     conda config --add channels conda-forge --force
     conda config --set channel_priority strict
     conda config --set safety_checks disabled
+    conda create --name TEST python=$PY --file requirements.txt --file requirements_dev.txt --file requirements_docs.txt --file requirements_plus.txt
```
ocefpaf: This is the "core" of this PR, it will create the environment for each python in the matrix based on the same files used in `setup.py`, so the conda test env and a pip installation are installing the same packages.

.travis.yml
```diff
@@ -18,17 +23,22 @@ install:
...
-  - conda env create --file="${ENV_FILE}"
-  - source activate test
-  - conda list
-  - conda info --all
+install:
+  - pip install -e . --no-deps --force-reinstall
 
 script:
-  - python setup.py sdist >/dev/null
-  - python setup.py install
```
ocefpaf: The `tarball` test substitutes `python setup.py sdist` and `pip install -e . --no-deps --force-reinstall` is better then `python setup.py install` b/c it won't try to pull the dependencies and uses `pip`, as recommended by PyPA. (`python setup.py install` will be removed at some point.)

requirements_dev.txt
```diff
@@ -7,3 +7,5 @@ coveralls
 bokeh
 ipywidgets
 statsmodels
+twine
+wheel
```
ocefpaf: Used in the tarball check.

requirements_docs.txt
```diff
@@ -1,6 +1,6 @@
 sphinx>=1.4.3
 sphinxcontrib-napoleon
-sphinx_gallery
+sphinx-gallery
```
ocefpaf: `pip` does not care about `-` or `_` but conda is unambiguous and accepts only one. This now works with both `conda` and `pip`.

requirements_plus.txt
```diff
@@ -7,6 +7,7 @@ geojson>=1.3.2
 folium>=0.2.1
 mplleaflet>=0.0.5
 statsmodels>=0.6.1
+mapclassify
```
ocefpaf: This is the only dependency I ported from the env files. However, I'm not familiar with the `pysal` code and maybe the tests require more packages here. (Locally I noticed that 49 tests were skipped.)

setup.py
```diff
@@ -60,6 +60,7 @@ def setup_package():
         version=__version__,
         description="A library of spatial analysis functions.",
         long_description=long_description,
+        long_description_content_type="text/x-rst",
```
ocefpaf: Fixed a warning from the tarball check.

PS: I noticed that there are two README, a markdown and a ResT. PyPI now accepts Markdown, so if you want to join them into one I can update this PR or send a new one later.

sjsrey: That would be great!

.travis.yml
sjsrey: Good to know.

sjsrey:
Thanks for the detailed explanation. This is really helpful!

PR #1134, opened 2019-08-02
Rtdmove

sjsrey:
This will move pysal-meta off readthedocs and onto github pages. 

Keeping the `doc` directory around for now. 

Added `docsrc` where the building happens, and `docs` where static files live that are served on gh.


docsrc/Makefile
```diff
@@ -0,0 +1,31 @@
...
+github:
+	@make html
+	@cp -a _build/html/. ../docs
+
+
+
+clean:
+	rm -rf $(BUILDDIR)/*
+	rm -rf auto_examples/
+	rm -rf ../docs/*
```
weikang9009: Should this line be removed?

sjsrey: It cleans out possibly stale static files before a build.
Maybe this should be replaced with an rsync and deletion flag

weikang9009: So `make clean` before `make github`? 
If that is the case, might make sense to append a cleaning line at the end of  `github`:
`github:
	@make html
	@cp -a _build/html/. ../docs
        @rm -rf $(BUILDDIR)/*
`
to avoid pushing duplicating built html files (in docsrc/(BUILDDIR)/* and /docs) to github by `git add .` .

PR #1151, opened 2020-01-21
Update issue templates

jGaboardi:
This PR adds an issue template for bug reports, etc.


jGaboardi:
This seems to already be in our repo, but only initializes "feature request" when creating an issue...
closing for now.

Issue #1160, opened 2020-02-17
api breakage in examples


sjsrey:
As @darribas [noted](https://gitter.im/pysal/pysal?at=5e4af88448ca2e297458e201)  `from libpysal.examples import get_path` broke with 2.2.

PR #1161, opened 2020-02-17
BUG: fix api breakage #1160

sjsrey:
Maintain api in the meta package to address #1160.


knaaptime:
this will need to be updated slightly to include `tobler.model`

Issue #1172, opened 2020-04-20
wrong environment file


knaaptime:
the top level [environment.yml](https://github.com/pysal/pysal/blob/master/environment.yml) defines the doc dependencies rather than the overall package dependencies

jGaboardi:
@knaaptime This was addressed in #1175, correct?

knaaptime:
yep

PR #1175, opened 2020-04-27
fix for tobler refactor

knaaptime:
this addresses https://github.com/pysal/tobler/issues/66 (and #1172) by removing the deprecated `tobler.data` module (and fixing the pysal environment.yml)


sjsrey:
Thanks!

PR #1180, opened 2020-05-23
Lazy imports and addition of access package

sjsrey:
This starts the 2.3.0 release target


knaaptime:
need to add access to `envioronment.yml`

pysal/frozen.py
```diff
@@ -2,19 +2,11 @@
...
-    "spglm": "1.0.7",
-    "spint": "1.0.6",
-    "spreg": "1.0.4",
-    "spvcm": "0.3.0",
-    "tobler": "0.2.0",
-    "mapclassify": "2.2.0",
-    "splot": "1.1.2"
-}
\ No newline at end of file
+frozen_packages = {"libpysal": "4.2.2", "esda": "2.2.1",
```
jGaboardi, line 5: Will these (and those in `environment.yml` be updated to currently released packages? On first glance the versions of `libpysal` and `spaghetti` are not the latest.

sjsrey: Yes.

martinfleis:
@sjsrey the whole readme is now lowercase. I assume that was not an intention?

sjsrey:
> @sjsrey the whole readme is now lowercase. I assume that was not an intention?

Good catch! No, that wasn't the intention. I will have to see what automated that?

jGaboardi:
[Appveyor failing](https://ci.appveyor.com/project/pysal/pysal/builds/34053198/job/j9o496rfgejpjw9u) with:

```
1388    C:\Miniconda36-x64\envs\test\lib\site-packages\access-1.1.1-py3.6.egg\access\weights.py:91
1389    C:\Miniconda36-x64\envs\test\lib\site-packages\access-1.1.1-py3.6.egg\access\weights.py:91: DeprecationWarning: invalid escape sequence \s
1390    """
```

ljwolf:
On this, tab completion still works on my end. However, the API is a bit polluted now. 

```python
>>> import pysal
>>> dir(pysal)
['__builtins__',
 '__cached__',
 '__doc__',
 '__file__',
 '__loader__',
 '__name__',
 '__package__',
 '__path__',
 '__spec__',
 '__version__',
 'base', # this is new, not in the hierarchy
 'explore',
 'federation_hierarchy', # this is also new
 'lib',
 'memberships', # This is also new
 'model',
 'versions', # this is also new
 'viz']
```
I'd think `versions` -> `__version__` or `__subpackage_versions__`, with federation hierarchy and memberships staying in `base`. Is this possible?

sjsrey:
After a bit more introspection, the tab completion is a little awkward for the layers. It works fine for the packages once they're imported from the layers, but there is a bit of a lag completing on the layer name. In addition it does pollute the namespace so I think we can roll back for now and revisit it after the 2.3 release.

pysal/base.py
```diff
@@ -20,6 +20,7 @@
                'pointpats': 'explore',
                'inequality': 'explore',
                'spaghetti': 'explore',
+               'access': 'access',
```
weikang9009, line 23: Should this be ```'access': 'model',```?

sjsrey: Yes - good catch!

sjsrey:
Replacing this with #1189

Issue #1205, opened 2021-01-18
remove codecov


sjsrey:
We are [testing the imports](https://github.com/pysal/pysal/blob/master/pysal/tests/test_imports.py), but codecov isn't designed to [see this](https://github.com/pysal/pysal/pull/1204#partial-pull-merging).

In this case, codecov is more annoying than helpful. I think we should remove it from the meta package unless someone can figure out a way to make it more useful.

knaaptime:
cant see that coverage is a useful stat for the metapackage. I say we toss it

jGaboardi:
Do we want to simply turn off CodeCov or remove completely?

* Turning off can be done like [this](https://community.codecov.io/t/how-to-temporarily-turn-off-codecov-on-a-github-repository/942).
* Removing altogether is a bit more involved:
  * [ ] remove integration (*Settings->Intergrations->Configure->Repository access->X*)
  * [ ] remove `codecov.yml`
  * [ ] remove from requirements, and environments (including `ci/*`)
  * [ ] remove action from `unittests.yml`
  * [ ] other?

sjsrey:
Turning off makes sense to me.

jGaboardi:
This will have to done by someone with full admin privileges.

sjsrey:
i deselected pysal on the repos in our organization account that codecov reports. I'm assuming that takes care of it?

jGaboardi:
I would think so. Let's try a toy PR to check?

sjsrey:
https://github.com/pysal/pysal/pull/1206

jGaboardi:
resolved in #1207

PR #1207, opened 2021-01-18
removing codecov action & streamlining CI

jGaboardi:
This PR addresses #1205.


jGaboardi:
@sjsrey This appears to do the trick.

sjsrey (approved):

PR #1230, opened 2021-10-02
ENH: Adding momepy to explore

sjsrey:


requirements.txt
```diff
@@ -7,6 +7,7 @@ pointpats>=2.2.0
 segregation>=2.0.0
 spaghetti>=1.6.2
 mgwr>=2.1.2
+momepy>-0.5.0
```
jGaboardi, line 10: should this be `>=`?

jGaboardi:
License and copyright still needs to be worked out for `momepy` (see pysal/momepy#305).

ci/38.yml
```diff
@@ -34,7 +34,7 @@ dependencies:
   - segregation
   - pointpats
   - giddy
-  - mapclassify
```
martinfleis, line 37: Did you mean to remove `mapclassify` here?

pysal/explore/momepy/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+from momepy.distribution import *
+from momepy.diversity import *
+from momepy.elements import *
+from momepy.graph import *
+from momepy.intensity import *
+from momepy.shape import *
+from momepy.utils import *
+from momepy.weights import *
+from momepy.preprocessing import *
+from momepy.coins import *
```
martinfleis, line 12: I think that this is not necessary. `momepy` exposes all public function and classes in the main namespace, so it may be easier (and future-proof) to just

```py
from momepy import *
```

ci/38.yml
sjsrey: > Did you mean to remove `mapclassify` here?

yes, there was a duplicate of mapclassify in the spec.

sjsrey:
> License and copyright still needs to be worked out for `momepy` (see [pysal/momepy#305](https://github.com/pysal/momepy/issues/305)).

I don't think this holds up this pr as we are not releasing momepy src as part of the meta package. the former is essentially a dependency of the latter.

jGaboardi:
> > License and copyright still needs to be worked out for `momepy` (see [pysal/momepy#305](https://github.com/pysal/momepy/issues/305)).
> 
> I don't think this holds up this pr as we are not releasing momepy src as part of the meta package. the former is essentially a dependency of the latter.

This was more a note to keep it on our radar 😄

jGaboardi (approved):

PR #1240, opened 2022-01-16
2.6.0rc1

sjsrey:


requirements.txt
```diff
@@ -1,21 +1,21 @@
...
+libpysal>=4.5.1.post2
 access>=1.1.3
 esda>=2.4.1
 giddy>=2.3.3
 inequality>=1.0.0
 pointpats>=2.2.0
-segregation>=2.0.0
-spaghetti>=1.6.2
+segregation>=2.1.0
+spaghetti>=1.6.4
```
jGaboardi, line 8: `spaghetti` should be [`1.6.5`](https://github.com/pysal/spaghetti/releases/tag/v1.6.5)

requirements.txt
```diff
@@ -1,21 +1,21 @@
...
 mgwr>=2.1.2
-momepy>=0.5.0
+momepy>=0.5.1
```
jGaboardi, line 10: `momepy` should be [`0.5.2`](https://github.com/pysal/momepy/releases/tag/v0.5.2)

requirements.txt
sjsrey: this was frozen when spaghetti was 1.6.2. I want to make sure things work on the last freeze before bumping another rc, which I will do later this week/next weekend.

jGaboardi: Got it. Same for `momepy`?

requirements.txt
sjsrey: see above for spaghetti

jGaboardi (approved):

PR #1243, opened 2022-01-23
2.6.rc2

sjsrey:


CHANGELOG.md
```diff
@@ -1,3 +1,208 @@
...
+<a name="contributors"></a>
+## Contributors
+
+Many thanks to all of the following individuals who contributed to this release:
+
+
+ - Charles Schmidt
+ - Eli Knaap
+ - Elliott Sales De Andrade
+ - Gegen07
```
jGaboardi, line 198: The @gegen07 GH handle is Germano Barcelos. Should we remove the GH handle in favor of only his name or keep both?

jGaboardi (approved):

PR #1245, opened 2022-01-30
build: add release gh action

sjsrey:


martinfleis:
I think that this should work. At least the PyPI upload. I am not familiar with the GitHub release action used here, in geopandas we are using the official `actions/create-release`.

You will only need to add a secret called `PYPI_PASSWORD` containing PyPI access token.

.github/workflows/release_and_publish.yml
```diff
@@ -0,0 +1,53 @@
...
+         with:
+           python-version: "3.x"
+ 
+       - name: Install Dependencies
+         run: |
+           python -m pip install --upgrade pip
+           pip install setuptools wheel twine jupyter urllib3 pandas pyyaml
+           python setup.py sdist bdist_wheel
+ 
+       - name: Get the tag name
```
jGaboardi, line 37: These two steps ('Get the tag name' and 'Release') do work. However, in [the latest `spaghetti` release](https://github.com/pysal/spaghetti/releases/tag/v1.6.5) (and other submodules) we have condensed down to a single step using [`actions/github-script@v5`](https://github.com/pysal/spaghetti/blob/1fc9647beed70a470a5d0617436d63ecb6d79020/.github/workflows/release_and_publish.yml#L40-L48). This generates customizable Release Notes in a [`.github/release.yml`](https://github.com/pysal/spaghetti/blob/main/.github/release.yml) and the actual release. I would recommend we go with that solution, but will defer to your judgment.

You can see more examples of the generated Release Notes in [`jGaboardi/tigernet`](https://github.com/jGaboardi/tigernet/releases) v2.{7,8,9}.

jGaboardi:
@sjsrey So do you only want the release and pushed to PyPI? I just saw your [message on gitter](https://gitter.im/pysal/pysal?at=61f6bfc3e1a1264f0a80838b).

sjsrey:
I want to do the release notes manually rather than as part of the action.

jGaboardi (approved):

jGaboardi:
:shipit:

PR #1247, opened 2022-01-30
update docs for 2.6

sjsrey:


docs/_static/documentation_options.js
```diff
@@ -1,10 +1,11 @@
 var DOCUMENTATION_OPTIONS = {
     URL_ROOT: document.getElementById("documentation_options").getAttribute('data-url_root'),
-    VERSION: '2.4.0',
+    VERSION: '2.2.0+217.geefab596.dirty',
```
jGaboardi, line 3: Still seeing the `dirty` version, but this we be cleared up with the next, push, correct? 

@martinfleis

martinfleis: @sjsrey you need to merge master into this branch and run sphinx again to get the correct version.

knaaptime:
i think this can be trimmed down a bit. Since the docs build in the cloud, we dont need any of the generated files inside the repo, just the source rsts

docs/_static/documentation_options.js
knaaptime: i think this is just a remnant from the last time the docs were build locally on serges machine, this gets overwritten in the gh-pages branch when the docs are built as part of GHA

knaaptime:
do a `make clean` and push again

sjsrey:
after cutting the release and fetching, when I make clean; make html I'm getting (locally):

image

knaaptime:
yeah maybe im wrong about this. We get the same ugliness on the live segregation docs at pysal.org/segregation

sjsrey:
That was due to work on the docs happening after the release commits.

sjsrey:
Maybe the fix is to publish out of gh-pages branch rather than master (which pysal is doing now). this way when we push to master the docs are not rebuild everytime. we update the docs only when pushing to gh-pages branch?

sjsrey:
> Maybe the fix is to publish out of gh-pages branch rather than master (which pysal is doing now). this way when we push to master the docs are not rebuild everytime. we update the docs only when pushing to gh-pages branch?

I thing gpd is wrestling with the same:

image

sjsrey:
For now, I'm ok with the docs pointing to the latest master. (Rather than the -2 version they are pointing to now). So I would say we merge this, then have a dev discussion about possible approaches to the docs and versioning across the ecosystem (on Friday).

martinfleis:
The `.dirty` issue is something we haven't figured out in geopandas. We use the same RTD settings for geopandas, xyzservices and momepy and the issue shows up only in geopandas. No clue why.

knaaptime:
is geopandas actually using RTD? we're only using gh-pages here

i think it may have to do with whether you leave the pre-built version of the docs in the repo or not. When i cooked up the gh-pages recipe about a year ago, the goal was to get rid of the docs/docsrc combo, so i was also trying to be as sparse as possible in the docs dir (i.e. i try to make sure to do a `make clean` before pushing so that the files generated by sphinx arent included. The GHA script runs after the commit, so it generates all those extra files in `_build` before publishing up to gh-pages. From git's perspective, those additional files mean that the git history is dirty because the generated files arent part of the last commit

knaaptime:
my guess, anyway

knaaptime:
...which makes me realize we might want to include the doc building [action](https://github.com/knaaptime/segregation/blob/master/.github/workflows/build_docs.yml) in momepy, which would ensure that the docs are published at pysal.org/momepy for consistency

martinfleis:
> is geopandas actually using RTD? we're only using gh-pages here

Yes, we no longer use `gh-pages`, everything is RTD now, meaning that there's nothing pre-built (the build folder is in `.gitignore`). But that is the same across those repos I mentioned.

> ...which makes me realize we might want to include the doc building action in momepy, which would ensure that the docs are published at pysal.org/momepy for consistency

as a duplication or instead of `docs.momepy.org`? I like the dedicated domain :).

knaaptime:
cool, your call obviously. I just default to consistency across the subpackages

martinfleis:
We can try making an alias, to have both.

knaaptime:
i think the issue is [here](https://github.com/pysal/momepy/blob/0a3d34da77faa63babcd2015e71f71c4b88d768c/.gitignore#L69)

momepy also includes `doc/generated` in gitignore but geopandas doesnt. That gets created by sphinx resulting in the dirty tag

martinfleis:
Hmm, not sure. GeoPandas has the equivalent [here](https://github.com/geopandas/geopandas/blob/a80081e52ed3bd069a75a4caaed44e45e35c647d/.gitignore#L65). The generated docs are at a different place. If you try running `make html` locally, git does not see any change.

One difference I realised now is that `momepy` and `xyzservices` are not using conda on RTD. That may be the key difference.

martinfleis:
Here - https://github.com/readthedocs/readthedocs.org/issues/8201

knaaptime:
nice find

knaaptime:
aside, is there a reason you guys prefer RTD to gh-pages? just the versioning feature? I ditched the former as soon as possible because its got these awkward fragility issues

martinfleis:
I don't think we ever had this discussion properly. Surely not recently in the era of GHA. But it is true that it can be tricky for geospatial packages. That is a reason why momepy notebooks are not executed on `make html` but are pre-executed. I just wasn't able to fit under the RTD memory limit, so momepy might be another candidate for moving to gh-pages. With geopandas we haven't had issues recently so there was no need to think about it.

PR #1265, opened 2022-07-23
bump requirements

sjsrey:


jGaboardi:
Is it too late to include a fresh release of `spreg`? Yesterday @pedrovma asked me to cut a new one. After some infrastructure updates in pysal/spreg#99 it will be ready to go.

sjsrey:
> 

I think we should hold off as we want to stick with the policy of freeze july 1?

jGaboardi:
> I think we should hold off as we want to stick with the policy of freeze july 1?

Ah! I got my months confused. I thought it was an August 1st freeze :face_in_clouds:

pedrovma:
Thank you so much, @jGaboardi ! My bad.

PR #1283, opened 2023-08-01
update changelog for 23.07

sjsrey:


knaaptime (commented):
😎

jGaboardi:
Looks like I forgot to get an `spopt` release out in time... :grimacing:

knaaptime:
if you cut one now, everyone still gets the latest and greatest if they `mamba install pysal` :)

jGaboardi:
Haste makes waste. Next time I will make a note to do it before the deadline.

Issue #1301, opened 2024-02-19
Pin on dependencies


sjsrey:
Update pins for packages.
[ref](https://discord.com/channels/1192517762103398401/1192517763986632766/1206989155045998622)

jGaboardi:
Closed via https://github.com/pysal/pysal/pull/1302

PR #1302, opened 2024-02-19
pin packages in requirements.txt

sjsrey:
Addresses #1301 

The build scripts did update requirements.txt, only I have not been adding that file in the commits.
We are working on fully automating the build, so this kind of pilot error will be avoided in the future.


jGaboardi:
Though not critical or urgent (or in this PR), we should at some point soon probably also:
* migrate to the condensed `pyproject.toml` schema over the `setup.py`+`setup.cfg`+`requirements*.txt`+etc.
* swap from `versioneer` to `setuptools_scm`

jGaboardi (approved):

martinfleis (approved):

Issue #1299, opened 2024-01-30
update `unittests.yml` workflow


jGaboardi:
[`unittests.yml`](https://github.com/pysal/pysal/blob/main/.github/workflows/unittests.yml) needs a healthy update.

PR #1305, opened 2024-05-12
ci update

sjsrey:
Addressing #1299


sjsrey:
This was modeled after [spopt ci](https://github.com/pysal/spopt/blob/main/.github/workflows/testing.yml), for reference.

.github/workflows/testing.yml
```diff
@@ -0,0 +1,87 @@
...
+       
+       - name: spatial versions
+         run: 'python -c "import geopandas; geopandas.show_versions();"'
+       
+       - name: run tests
+         run: |
+           pytest pysal \
+           -v \
+           -r a \
+           -n logical \
```
jGaboardi, line 79: In order for multicore testing to work, we need to add `pytest-xdist` to all `ci/` environments.

ci/310-oldest.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - giddy
+  - mapclassify
+  - seaborn
+  - descartes
+  - numba
+  - momepy
+  - inequality
+  - access
+  - splot
+  - pip:
```
jGaboardi, line 58: Shall we move `spopt` into the `dependencies` section (in all environments)?

ci/310-oldest.yaml
jGaboardi, line 1: For the `-oldest` tag we should ideally have minimum pins. We could probably simply [follow SPEC000](https://scientific-python.org/specs/spec-0000/) for this.

.github/workflows/testing.yml
```diff
@@ -0,0 +1,87 @@
...
+         with:
+           fetch-depth: 0 # Fetch all history for all branches and tags.
+       
+       - name: setup micromamba
+         uses: mamba-org/setup-micromamba@v1
+         with:
+           environment-file: ${{ matrix.environment-file }}
+           micromamba-version: 'latest'
+        
+       ##############################################################################
```
jGaboardi, line 59: We can remove this section for install `pulp`

ci/312-dev.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - scikit-learn
+    - scipy
+    - git+https://github.com/geopandas/geopandas.git
+    - git+https://github.com/networkx/networkx.git
+    - git+https://github.com/shapely/shapely.git
+    - git+https://github.com/pysal/libpysal.git
+    - git+https://github.com/pysal/esda.git
+    - git+https://github.com/pysal/giddy.git
+    - git+https://github.com/pysal/inequality.git
+    - git+https://github.com/pysal/momempy.git
```
jGaboardi, line 35:
```suggestion
    - git+https://github.com/pysal/momepy.git
```

.github/workflows/testing.yml
```diff
@@ -0,0 +1,87 @@
...
+            ci/310-latest.yaml,
+            ci/311-latest.yaml,
+            ci/312-latest.yaml,
+            ci/312-dev.yaml,
+         ]
+         include:
+           - environment-file: ci/312-latest.yaml
+             os: macos-latest
+           - environment-file: ci/312-latest.yaml
+             os: macos-14 # Apple Silicon
```
martinfleis, line 38:
```suggestion
             os: macos-latest # Apple Silicon
           - environment-file: ci/312-latest.yaml
             os: macos-13 # Intel
```

GHA default is being migrated.

jGaboardi:
Looks like [`pysal` isn't actually being installed](https://github.com/pysal/pysal/actions/runs/9055028425/job/24875540666?pr=1305#step:6:28) here?

```
_________________ ERROR collecting pysal/tests/test_imports.py _________________
ImportError while importing test module '/home/runner/work/pysal/pysal/pysal/tests/test_imports.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../../micromamba/envs/test/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
pysal/tests/test_imports.py:2: in <module>
    from pysal import federation_hierarchy
E   ModuleNotFoundError: No module named 'pysal'
```

jGaboardi:
***re: pacakage install***

For some packages (e.g. `spopt` and [`spaghetti`](https://github.com/pysal/spaghetti/blob/main/.github/workflows/testing.yml)), merely [checking out the repo is enough](https://github.com/pysal/spaghetti/blob/17a411cb88dfe1532e01b83bdb0cca28cdd926a5/.github/workflows/testing.yml#L46). However, this is not the case most of time (e.g. [`tobler`](https://github.com/pysal/tobler/blob/1439e4268ec18df382ba1f5a78dcec578639a5ae/.github/workflows/unittests.yml#L65) and [`momepy`](https://github.com/pysal/momepy/blob/36b13f827200c694ecbcff54545b227d526959b6/.github/workflows/tests.yaml#L53)). I have never been able to figure the difference and why it is not needed sometimes. Seems like it is needed here.

.github/workflows/testing.yml
jGaboardi: @martinfleis Is [this](https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/) the news?

knaaptime (approved):

ci/310-oldest.yaml
martinfleis: That is a bit tricky to do in here as you would need to pin 2nd order dependencies, given the pysal depends only on subpackages, where we always want to pull the latest, never the oldest. I would probably leave this env out.

jGaboardi: @sjsrey If we follow @martinfleis's idea here, we can simply rename `310-oldest.yaml` --> `310-latest.yaml` (and in the workflow file) and have done with it.

ci/311-latest.yaml
jGaboardi, line 1: Is this file still even needed? I opened #1306 because I thought it could be pruned entirely.

ci/310-oldest.yaml
jGaboardi: Gentle ping to remind to remove the `pip: spopt` and move `spopt` directly up beneath `splot` in an CI environments for this PR.

PR #1337, opened 2024-07-10
removing `spvcm` from meta release -- archived - #1330

### LINKED ISSUE #1330: inclusion of `spvcm` in PySAL meta-release
https://github.com/pysal/spvcm/issues/17#issuecomment-2217980601

> 3. As for distribution with `pysal/pysal`...
>    1. Should we:
>       1. ... update the README, docs, landing page, etc. that `spvcm` will not be distributed in the next meta release?
>       2. ... everything from (1.) + immediately remove from [`pysal/pysal/pyproject.toml`](https://github.com/pysal/pysal/blob/12ae3c1a59b02e3142d9f9e083c211e8b2fdc314/pyproject.toml#L51)

xref:
* #1322
* #1324

jGaboardi:
This PR removing `spvcm` from meta release, as it is now archived - https://github.com/pysal/spvcm/issues/18
* resolves #1330


sjsrey:
I'm not sure whether we deprecated spvcm before, so removing it from meta may be pretty drastic in case we haven't. If we haven't, we should deprecate spvcm during the 24.07 meta cycle with documentation, then remove in 25.01.

knaaptime:
dunno either, but also not sure how would we issue a deprecation warning in meta?

jGaboardi:
OK, so I shouldn't have arched it like we discussed? Should I "unarchive"?

sjsrey:
> OK, so I shouldn't have arched it like we discussed? Should I "unarchive"?

No, archiving is fine, I don't think we need to revert that.

I'm wondering about the best way to surface the archiving in the meta release.

sjsrey:
What about updating the meta README.md like:

image

jGaboardi:
Works for me.

sjsrey:
So we would keep this pr open until 25.01 release, but I can add the warning to the docs in the 24.07 release.

jGaboardi:
> So we would keep this pr open until the 25.01 release, but I can add the warning to the docs in the 24.07 release.

Probably the best idea. --> Should also post a notice like I suggest in https://github.com/pysal/pysal.github.io/issues/229?

sjsrey (approved):

jGaboardi:
@sjsrey Let's go ahead with merge here?

Issue #1347, opened 2024-07-21
Missing dependency for doc build


sjsrey:
[sphinx-design](https://github.com/pysal/pysal/actions/runs/10031995653/job/27723093047#step:5:14) is missing.

`sphinx-design` is listed in the [pyproject.toml](https://github.com/pysal/pysal/actions/runs/10031995653/job/27723093047#step:5:14) file.
Do we need to install it otherwise?

jGaboardi:
It needs to also be included in [`ci/312-latest.yaml`](https://github.com/pysal/pysal/blob/e40a6339fd4e5d63380feb8dcac841ffc629aa5d/ci/312-latest.yaml#L59-L64) -- see [here](https://github.com/pysal/pysal/blob/e40a6339fd4e5d63380feb8dcac841ffc629aa5d/.github/workflows/build_docs.yml#L22) in `.github/workflows/build_docs.yml`

jGaboardi:
> sphinx-design is listed in the [pyproject.toml](https://github.com/pysal/pysal/actions/runs/10031995653/job/27723093047#step:5:14) file.

In order to install optional dependencies from `pyproject.toml` you would have to explicitly list the associated group. For example, to install [`plus` dependencies](https://github.com/pysal/pysal/blob/e40a6339fd4e5d63380feb8dcac841ffc629aa5d/pyproject.toml#L63):

```
$ pip install .[plus]
```

sjsrey:
Closing thanks to @jGaboardi

PR #1348, opened 2024-07-21
BUG: install sphinx_design for gh workflow Build Docs

sjsrey:
#1347

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -39,8 +39,8 @@ jobs:
 
       - name: Install
         run: |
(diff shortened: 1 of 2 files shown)


jGaboardi:
We *could* do it this way, but better to simply add `sphinx_design` into `ci/312-latest.yaml` because we should also remove `sphinx_bootstrap_theme` from there.

jGaboardi (approved):

jGaboardi:
Let's stick with `conda-forge` --> simply needs to have a hyphen, not underscore: [`sphinx-design`](https://anaconda.org/conda-forge/sphinx-design)

PR #1357, opened 2024-09-14
pin fiona in oldest

jGaboardi:
This PR
* pins `fiona<1.10` in the the `310-oldest` CI environment
* cleans up & organizes CI environments

### DIFF
--- a/ci/310-latest.yaml
+++ b/ci/310-latest.yaml
@@ -3,59 +3,50 @@ channels:
   - conda-forge
 dependencies:
   - python=3.10
-  - six
-  - pyyaml
-  - deprecated
-  # required
-  # testing
-  - matplotlib
-  - pytest
-  - pytest-cov
-  - coverage
-  - codecov
-  # optional
+  # ecosystem
+  - access
+  - esda
+  - giddy
+  - inequality
+  - libpysal
+  - mapclassify
+  - mgwr
+  - momepy
+  - pointpats
+  - segregation
+  - spaghetti
+  - splot
+  - spint
+  - spopt
+  - spreg
+  - spvcm
+  - tobler>=0.2.1
+  # external & optiopnal
   - bokeh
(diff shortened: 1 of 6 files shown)


martinfleis:
Somehow it is not strictly needed here as our tests in here do not invoke that incompatible code. 

Also, no clue what that error is.

jGaboardi:
> Somehow it is not strictly needed here as our tests in here do not invoke that incompatible code.

Good to have for conformity though, no?

> Also, no clue what that error is.

That was due to a `yaml` syntax error

ci/310-latest.yaml
```diff
@@ -3,59 +3,50 @@ channels:
...
+  - spopt
+  - spreg
+  - spvcm
+  - tobler>=0.2.1
+  # external & optiopnal
   - bokeh
-  - scikit-learn
+  - deprecated
+  - dill
+  - fiona
```
martinfleis, line 28: Do we need Fiona at all?

martinfleis: I mean, in the latest. We need that pin in the oldest.

jGaboardi: suppose not?

gegen07 (approved):
Why we should pin fiona to 1.10 in oldest? Maybe I am missing something important.

jGaboardi:
> Why we should pin fiona to 1.10 in oldest? Maybe I am missing something important.

See [here](https://github.com/pysal/momepy/pull/654). Has to do with older `geopandas` mixing with latest `fiona`

martinfleis:
@gegen07 we're pinning old geopandas which is not compatible with the latest Fiona. If we don't pin Fiona, we hit an error on missing `fiona.path` module. It is more important in CI envs where this actually happens than here.

knaaptime (approved):

PR #1364, opened 2025-01-31
25.01

sjsrey:
This adds the changelog.md file and a few updates,.


sjsrey:
@knaaptime the changelog.md file is not created by the action but rather locally. We should revisit this to streamline the process. For now, if we merge this, we can rerun the release action. Then, we can turn to more streamlining after the release.

knaaptime:
ah ok. i'd forgotten but i think that's intentional... If we have it autogenerated during the commit, it's harder to manually add highlights

knaaptime:
do we want both changes.md and changelog.md in here?

sjsrey:
> do we want both changes.md and changelog.md in here?

looks like the action requires `changelog.md` so we don't need  `changes.md`

knaaptime (approved):

PR #1366, opened 2025-02-05
CI: new manual gh action for pypi release

sjsrey:
This isolates the push to pypi in cases where the gh release is already built.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,39 @@
+name: Publish to PyPI
+
+on:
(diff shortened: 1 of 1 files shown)


jGaboardi:
So the idea here is not to replace `release_and_publish.yml`, but to have an optional tag+publish that only runs manually, correct?

sjsrey:
Exactly. Some changes in the CI since 24.07 have led to the publish and docs actions failing.

And this might help debug.

xref #1367

jGaboardi (approved):

PR #1371, opened 2025-02-06
Correct pypi push action

sjsrey:


.github/workflows/publish.yml
```diff
@@ -31,9 +31,8 @@ jobs:
       - name: Build the package
         run: python -m build
 
-      - name: Publish to PyPI
-        env:
-          TWINE_USERNAME: __token__
-          TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
-        run: twine upload dist/*
-
+      - name: Publish distribution � to PyPI
```
jGaboardi, line 34:
```suggestion
      - name: Publish distribution to PyPI
```

jGaboardi (approved):

knaaptime:
before i lose this again, i forgot that you can also trigger a workflow from the github CLI https://cli.github.com/manual/gh_workflow_run

`The given workflow file must support an on.workflow_dispatch trigger in order to be run in this way.`

which might be useful here

jGaboardi:
@knaaptime -- good point. We have that on [`testing`](https://github.com/pysal/pysal/blob/e6db2d7dcd340e242459e8289578bfb4a4830298/.github/workflows/testing.yml#L12-L17) and [`build_docs.yml`](https://github.com/pysal/pysal/blob/e6db2d7dcd340e242459e8289578bfb4a4830298/.github/workflows/build_docs.yml#L7-L12) flows, but not here.

We should probably be adding a manual trigger to any actions. Can we think of any reason why not?

PR #1408, opened 2026-01-26
TEST: Add test for _installed_version with nonexistent package

nimra06:
## What This Does

Adds the missing test case for _installed_version() when called with a nonexistent package. The function correctly returns 'NA', but the test was never added (there was a comment saying it was skipped due to a bug that doesn't actually exist).

## Changes

- Added test_installed_version_for_nonexistent_package() method
- Removed outdated comment about NameError bug
- Test verifies function returns 'NA' for nonexistent packages

## Checklist

- [x] Tests pass locally
- [x] Code is formatted and linted
- [x] PR is directed to `pysal/main` branch

## Related

Closes #1407

PR #1419, opened 2026-02-07
Add test for _installed_version with non-existent package

AKSrivastava28:
This PR adds missing test coverage for _installed_version() when a non-existent
package name is provided, asserting that the function correctly returns "NA".

The function already behaves as expected; this change only adds a test and
removes outdated assumptions about a previously mentioned bug.


jGaboardi:
This is virtually the exact same PR as #1408. Please go through open PRs prior to starting new ones.

PR #1442, opened 2026-07-24
26.7rc1

sjsrey:
Update the release build tooling for 26.7rc1

The major change here in the meta package is we are no longer pinning the pysal subpackages.

This is because each of the 18 sub-packages already declares its own dependencies — including constraints on other PySAL packages. For example, giddy declares esda<2.9,>=2.7. Those inter-package constraints are the sub-packages' responsibility to maintain, and pip's dependency resolver is designed to honor them.

When pysal also declares esda>=2.10.0, it creates a second, conflicting constraint that pip has to satisfy simultaneously. Since giddy 2.3.8 requires esda<2.9, the two constraints are irreconcilable and installation fails — even though the sub-packages themselves are internally consistent.

By listing the 18 packages without version pins, pysal steps back and lets the resolver do its job: it finds a set of versions that satisfies all the sub-packages' own declared constraints. The result is always a self-consistent installation.

The downside of this is that some sub-package may not be at its latest release. For example, consider the situation with `giddy` and  `esda`:

- giddy 2.3.8 (latest) requires esda<2.9
- esda 2.10.0 (latest) exists

If a user runs pip install pysal, the resolver sees that it needs both giddy and esda. Since giddy constrains esda<2.9, the resolver installs esda 2.8.x — not the latest.

The meta-package will now give a consistent, working installation - but some sub-packages may not be at their latest release.


ci/311-oldest.yml
jGaboardi, line 1: This 3.11 environment file can be entirely removed, right?

sjsrey: Yes. spec0

https://github.com/pysal/pysal/pull/1442/changes/6f95b8af04a24327f87e84e1054af8574aecdd75

jGaboardi: But this file is still `3.11-oldest` and pinned to Python 3.11. Should it be renamed to `3.12-oldest` and pinned to Python 3.12?

sjsrey: I'm not seeing `3.11-oldest`?

sjsrey: remind me of the `-latest` vs. `-oldest` differentiation?

jGaboardi: This is what I am seeing, but maybe GitHub is not updating the diff correctly.

Screenshot 2026-07-24 at 5 39 20 PM

jGaboardi: > remind me of the `-latest` vs. `-oldest` differentiation?

* `-latest` environments pull the current latest available versions of dependencies
* `-oldest` environments provide some pins for the oldest supported dependency versions 

However, I seem to remember that we aren't that strict with `oldest` in meta-pysal since those oldest tests are kind of the responsibility of the federated packages themselves.

martinfleis:
If we remove the pins, what is the point of making releases? Given they will be the same...

giddy upper pins should be resolved in giddy and should not have been in place at all...

sjsrey:
I agree and would think we want the pins, but with the current release of giddy the meta package is broken with pinning. 
The solution is probably to cut a point release of giddy with the corrected pins.

jGaboardi (approved):

